PR #4, opened 2015-12-17
backtrace: add support for generic backtraces

kyrylo:
Sometimes exceptions can contain manual backtraces that quack similar to
Ruby exceptions, but do not stricly adhere their template. I stumbled
upon this issues, when I was working with SCSS. It sets a couple of
custom first frames such as:

```
Sass::SyntaxError: Undefined mixin 'animation'.
/home/app/assets/stylesheets/error_pages.scss:139:in `animation'
/home/app/app/assets/stylesheets/error_pages.scss:139
```

Previously, we were not able to parse the second frame.

Furthermore, we now raise error if we cannot parse a stack frame to
simplify debugguging and improving our parser.

### DIFF
--- a/lib/airbrake-ruby/backtrace.rb
+++ b/lib/airbrake-ruby/backtrace.rb
@@ -86,11 +86,10 @@ module Airbrake
       match = regexp.match(stackframe)
       return match if match
 
-      unless (matched_frame = GENERIC_STACKFRAME_REGEXP.match(stackframe))
-        raise Airbrake::Error, "failed parsing '#{stackframe}'"
-      end
+      match = GENERIC_STACKFRAME_REGEXP.match(stackframe)
+      return match if match
 
-      matched_frame
+      raise Airbrake::Error, "can't parse '#{stackframe}'"
     end
   end
 end
--- a/lib/airbrake-ruby/backtrace.rb
+++ b/lib/airbrake-ruby/backtrace.rb
@@ -14,7 +14,7 @@ module Airbrake
     ##
     # @return [Regexp] the pattern that matches standard Ruby stack frames,
     #   such as ./spec/notice_spec.rb:43:in `block (3 levels) in <top (required)>'
-    STACKFRAME_REGEXP = %r{\A
+    RUBY_STACKFRAME_REGEXP = %r{\A
       (?<file>.+)       # Matches './spec/notice_spec.rb'
       :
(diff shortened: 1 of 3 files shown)


lib/airbrake-ruby/backtrace.rb
```diff
@@ -71,5 +81,16 @@ def stack_frame(match)
           function: match[:function] }
       end
     end
+
+    def self.match_frame(regexp, stackframe)
+      match = regexp.match(stackframe)
+      return match if match
+
+      unless (matched_frame = GENERIC_STACKFRAME_REGEXP.match(stackframe))
+        raise Airbrake::Error, "failed parsing '#{stackframe}'"
```
vmihailenco: Nitpick: "Failed parsing" assumes that there was some failure, but here we just "can't parse".

lib/airbrake-ruby/backtrace.rb
```diff
@@ -71,5 +81,16 @@ def stack_frame(match)
...
+      end
+
+      matched_frame
```
vmihailenco: You have `match` and `matched_frame`. Also your function looks following

```
return match
raise error
return match
```

but would be nice if it looks like

```
return match
return match
raise error
```

vmihailenco:
I don't understand who is responsible to parse `Sass::SyntaxError: Undefined mixin 'animation'.` from your example...

lib/airbrake-ruby/backtrace.rb
kyrylo: Fixed by https://github.com/airbrake/airbrake-ruby/commit/b31d680b8e314fb5ab3915b92896f1827ce7927d

lib/airbrake-ruby/backtrace.rb
kyrylo: Fixed by https://github.com/airbrake/airbrake-ruby/commit/b31d680b8e314fb5ab3915b92896f1827ce7927d

kyrylo:
`Sass::SyntaxError` is an exception. `Undefined mixin 'animation'` is its message. 

Its backtrace:

```
/home/app/assets/stylesheets/error_pages.scss:139:in `animation'
/home/app/app/assets/stylesheets/error_pages.scss:139
<other frames were omitted>
```

vmihailenco:
LGTM

PR #5, opened 2015-12-17
notice: specify the language clearer in the context payload

kyrylo:


vmihailenco:
Do we care about Ruby in JVM and other flavors? E.g. it could be JRuby/2.2.2

kyrylo:
It's trivial to add support for this, so I don't see why not :)
Added & squashed commits.
PTAL

vmihailenco:
Would be nice to have a test.

LGTM

kyrylo:
The test wouldn't be trivial, because we run the test suite against many Rubies and its versions.

PR #11, opened 2015-12-22
Fix OS match regex

zefer:
The previous regex failed to match OS values containing a `.` such as
`x86_64-darwin14.0`, causing the test to fail on such systems.


kyrylo:
LGTM

PR #14, opened 2015-12-25
airbrake-ruby: ignore SystemExit in the `at_exit` hook

### LINKED ISSUE #13: Unnecessary airbrake error when running unknown rake task
* Airbrake version: 5.0.1
* Ruby version: 2.2.3
* Framework name & version: Rails 4.2.5

#### Airbrake config

````
    config.project_id = <project_id>
    config.project_key = <project_key>
    config.host = <host>
    config.ignore_environments = [:test]
  end
````

#### Description

I checked that my config is OK by running `rake airbrake:test`, it runs successfully.

But when I run a rake task that does not exist I am getting the following output:

````
$ bin/rake kokoko:dskfkl
rake aborted!
Don't know how to build task 'kokoko:dskfkl'

(See full trace by running task with --trace)
/Users/vkhustochka/.rvm/gems/ruby-2.2.3@quails/gems/airbrake-ruby-1.0.1/lib/airbrake-ruby.rb:282:in `call_notifier': the 'default' notifier isn't configured (Airbrake::Error)
	from /Users/vkhustochka/.rvm/gems/ruby-2.2.3@quails/gems/airbrake-ruby-1.0.1/lib/airbrake-ruby.rb:149:in `notify_sync'
	from /Users/vkhustochka/.rvm/gems/ruby-2.2.3@quails/gems/airbrake-ruby-1.0.1/lib/airbrake-ruby.rb:291:in `block in <top (required)>'
````

I think the issue is that in this case airbrake is loaded (through bundler) but initializers are not loaded, so Airbrake config is not loaded.

### PR DESCRIPTION
**Pull request by kyrylo**
Fixes #13 (Unnecessary airbrake error when running unknown rake task)

The problem with Rake is that on unknown task it
[raises SystemExit][1] with help of [`Kernel.exit`][2], which we catch
and try to send to Airbrake, which results in an Airbrake::Error.

This is the easiest fix. I tried messing with the airbrake gem to load
the Airbrake config when you launch a Rake task inside a Rails app, but
it's not easy to do: the Rails app is not initialised at that moment.

I stopped digging there, because I was getting into the rabbit hole.

[1]: https://github.com/ruby/rake/blob/805c13ab52eae9df55f9030f40c7fbc1beadc105/lib/rake/application.rb#L191
[2]: http://ruby-doc.org/core-2.2.3/Kernel.html#method-i-exit

### DIFF
--- a/lib/airbrake-ruby.rb
+++ b/lib/airbrake-ruby.rb
@@ -286,7 +286,9 @@ module Airbrake
   end
 end
 
-# Notify of unhandled exceptions, if there were any.
+# Notify of unhandled exceptions, if there were any, but ignore SystemExit.
 at_exit do
-  Airbrake.notify_sync($ERROR_INFO) if $ERROR_INFO
+  if $ERROR_INFO && $ERROR_INFO.class != SystemExit
+    Airbrake.notify_sync($ERROR_INFO)
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ Airbrake Ruby Changelog
 
 ### master
 
+* Ignored `SystemExit` in the `at_exit` hook, which has fixed the Rake
+  integration for the [airbrake gem][airbrake-gem] gem
+  ([#14](https://github.com/airbrake/airbrake-ruby/pull/14))
+
 ### [v1.0.1][v1.0.1] (December 22, 2015)
 
 * Fixed the `Airbrake.add_filter` block API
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: lib/airbrake-ruby.rb [hunk 1]
```diff
@@ -286,7 +286,9 @@ def call_notifier(notifier, method, *args, &block)
   end
 end
 
-# Notify of unhandled exceptions, if there were any.
+# Notify of unhandled exceptions, if there were any, but ignore SystemExit.
 at_exit do
-  Airbrake.notify_sync($ERROR_INFO) if $ERROR_INFO
+  if $ERROR_INFO && $ERROR_INFO.class != SystemExit
```
zefer: Can you find a way to test this change?

kyrylo: I doubt there's an easy way. I'd prefer not to mess with processes. Luckily, this hook will not get any more complex.

zefer: Okay, thanks for looking. I'm that case, maybe you can improve the comment to explain why we're ignoring `SystemExit`?

zefer:
LGTM

PR #20, opened 2016-01-07
notice: factor out code into the NestedException class

kyrylo:
The Notice class is reaching its limit of 120 lines. This change
introduces NestedException - a class with a single responsibility,
which simplifies Notice.


vmihailenco:
LGTM

I would not bother too much with max number of lines limit...

kyrylo:
I like limitations. They make you think better. You probably know this guy: http://www.artlebedev.com/mandership/175/

PR #21, opened 2016-01-07
notifier: make sure generated notices always have a backtrace

kyrylo:
In theory it's possible to generate a backtrace that consists only of
internal frames. When this happens, then we send a notice without a
backtrace, because all the internal frames are filtered. When we see a
notice without a backtrace in the dashboard, it's incredibly hard to
debug it.

This change ensures that in case of an internal error we at least can
debug the failure.

### DIFF
--- a/lib/airbrake-ruby/notifier.rb
+++ b/lib/airbrake-ruby/notifier.rb
@@ -139,7 +139,13 @@ module Airbrake
     end
 
     def clean_backtrace
-      caller.drop_while { |frame| frame.include?('/lib/airbrake') }
+      caller_copy = Kernel.caller
+      clean_bt = caller_copy.drop_while { |frame| frame.include?('/lib/airbrake') }
+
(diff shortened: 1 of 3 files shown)


lib/airbrake-ruby/notifier.rb
```diff
@@ -139,7 +139,13 @@ def default_sender
     end
 
     def clean_backtrace
-      caller.drop_while { |frame| frame.include?('/lib/airbrake') }
+      bt = caller
```
vmihailenco: Such variables usually have word `save/copy` in their name :)

vmihailenco:
Waiting for build to pass.

lib/airbrake-ruby/notifier.rb
kyrylo: Renamed to `caller_copy`. Thanks!

kyrylo:
PTAL

vmihailenco:
LGTM

PR #26, opened 2016-01-13
notifier: respawn workers if they're all dead

kyrylo:
Potentially fixes https://github.com/airbrake/airbrake/issues/463
(falling back to sync delivery because there are no running async
workers)

I noticed I can reliably reproduce this message when I invoke the Rails
console and manually send a notice via `Airbrake.notify`. It happens
because by default Rails launches [Spring][1], which preloads the app
by using the `fork()` system call. And when that happens, as we know,
our worker threads die.

The fix is simply to respawn workers. It works reliably with the Rails
console and I assume it will fix the above mentioned issue.

[1]: https://github.com/rails/spring/

### DIFF
--- a/lib/airbrake-ruby/async_sender.rb
+++ b/lib/airbrake-ruby/async_sender.rb
@@ -14,9 +14,7 @@ module Airbrake
       @sender = SyncSender.new(config)
       @closed = false
       @workers = ThreadGroup.new
-
-      config.workers.times { @workers.add(spawn_worker) }
-      @workers.enclose
+      @pid = nil
     end
 
     ##
@@ -67,17 +65,32 @@ module Airbrake
     # went wrong.
     #
     # Workers are expected to crash when you +fork+ the process the workers are
-    # living in. Another possible scenario is when you close the instance on
-    # +at_exit+, but some other +at_exit+ hook prevents the process from
-    # exiting.
+    # living in. In this case we detect a +fork+ and try to revive them here.
+    #
+    # Another possible scenario that crashes workers is when you close the
+    # instance on +at_exit+, but some other +at_exit+ hook prevents the process
+    # from exiting.
     #
     # @return [Boolean] true if an instance wasn't closed, but has no workers
     # @see https://goo.gl/oydz8h Example of at_exit that prevents exit
     def has_workers?
+      return false if @closed
+
+      if @pid != Process.pid && @workers.list.empty?
+        @pid = Process.pid
+        spawn_workers
+      end
+
       !@closed && @workers.list.any?
     end
 
     private
+
+    def spawn_workers
+      @workers = ThreadGroup.new
+      @config.workers.times { @workers.add(spawn_worker) }
+      @workers.enclose
+    end
 
     def spawn_worker
       Thread.new do
--- a/lib/airbrake-ruby/notifier.rb
+++ b/lib/airbrake-ruby/notifier.rb
@@ -127,15 +127,13 @@ module Airbrake
     end
 
     def default_sender
-      if @async_sender.has_workers?
-        @async_sender
-      else
-        @config.logger.warn(
-          "#{LOG_LABEL} falling back to sync delivery because there are no " \
-          "running async workers"
-        )
-        @sync_sender
-      end
+      return @async_sender if @async_sender.has_workers?
+
+      @config.logger.warn(
+        "#{LOG_LABEL} falling back to sync delivery because there are no " \
+        "running async workers"
+      )
+      @sync_sender
     end
 
     def clean_backtrace
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Airbrake Ruby Changelog
   ([#25](https://github.com/airbrake/airbrake-ruby/pull/25))
 * Made sure that generated notices always have a backtrace
   ([#21](https://github.com/airbrake/airbrake-ruby/pull/21))
+* Made the asynchronous delivery mechanism more robust
+  ([#26](https://github.com/airbrake/airbrake-ruby/pull/26))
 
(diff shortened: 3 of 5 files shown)


lib/airbrake-ruby/async_sender.rb
```diff
@@ -13,10 +13,9 @@ def initialize(config)
       @unsent = SizedQueue.new(config.queue_size)
       @sender = SyncSender.new(config)
       @closed = false
-      @workers = ThreadGroup.new
+      @workers = nil
 
-      config.workers.times { @workers.add(spawn_worker) }
-      @workers.enclose
+      spawn_workers
```
vmihailenco: Remove this? It is kinda nice that we don't spawn any workers if async notifier is not used.

lib/airbrake-ruby/notifier.rb
```diff
@@ -126,7 +126,9 @@ def send_notice(exception, params, sender = default_sender)
       sender.send(notice)
     end
 
-    def default_sender
+    def select_sender
```
vmihailenco: I don't like new name - sounds like a SQL query.

lib/airbrake-ruby/notifier.rb
```diff
@@ -126,7 +126,9 @@ def send_notice(exception, params, sender = default_sender)
...
+      @async_sender.spawn_workers unless @async_sender.has_workers?
+
       if @async_sender.has_workers?
```
vmihailenco: So now you
- spawn workers
- immediately check if there are workers

Sounds wrong and dangerous if workers die sending notices:
1. spawn workers
2. send async
3. workers die and notices are not sent
4. go to step 1

vmihailenco:
Can you instead check that thread was forked and respawn workers?

lib/airbrake-ruby/async_sender.rb
kyrylo: Removed.

lib/airbrake-ruby/notifier.rb
kyrylo: Reverted.

lib/airbrake-ruby/notifier.rb
kyrylo: Refactored.

kyrylo:
PTAL

lib/airbrake-ruby/notifier.rb
```diff
@@ -42,6 +44,12 @@ def initialize(user_config)
     ##
     # @macro see_public_api_method
     def notify(exception, params = {})
+      if !@workers_spawned || @pid != Process.pid
```
vmihailenco: Do you really need `!@workers_spawned` check? I think `@pid != Process.pid` should be enough...

lib/airbrake-ruby/notifier.rb
```diff
@@ -42,6 +44,12 @@ def initialize(user_config)
...
+        @pid = Process.pid
+        @async_sender.spawn_workers
+        @workers_spawned = true
+      end
```
vmihailenco: Please move this block to `@async_sender.has_workers?`

lib/airbrake-ruby/async_sender.rb
```diff
@@ -41,6 +38,11 @@ def close
         raise Airbrake::Error, 'attempted to close already closed sender'
       end
 
+      unless @workers
```
vmihailenco: What this check achieves?

kyrylo: Covers the case when we create a notifier and close it immediately. If we do that, then `@workers == nil`, so we can return early and avoid failures about undefined methods.

kyrylo: That said, with the latest changes it's not needed anymore.

lib/airbrake-ruby/notifier.rb
kyrylo: Done.

lib/airbrake-ruby/notifier.rb
kyrylo: No, with the current refactoring we don't need it.

kyrylo:
PTAL

lib/airbrake-ruby/async_sender.rb
```diff
@@ -74,11 +72,22 @@ def closed?
     # @return [Boolean] true if an instance wasn't closed, but has no workers
     # @see https://goo.gl/oydz8h Example of at_exit that prevents exit
     def has_workers?
+      if @pid != Process.pid && @workers.list.empty?
+        @pid = Process.pid
+        spawn_workers
+      end
```
vmihailenco: I think you should not spawn any workers if notifier is closed.

kyrylo: Should this be enough, then?

```diff
diff --git a/lib/airbrake-ruby/async_sender.rb b/lib/airbrake-ruby/async_sender.rb
index 35b1aa4..dc3a71d 100644
--- a/lib/airbrake-ruby/async_sender.rb
+++ b/lib/airbrake-ruby/async_sender.rb
@@ -74,7 +74,7 @@ module Airbrake
     def has_workers?
       if @pid != Process.pid && @workers.list.empty?
         @pid = Process.pid
-        spawn_workers
+        spawn_workers unless closed?
       end
 
       !@closed && @workers.list.any?

```

vmihailenco: Why not

```
return false if @closed
```

kyrylo: Also works. My thinking was that in this case `@pid` won't be updated, but since the sender is closed it doesn't matter. Updated according to your suggestion.

kyrylo:
PTAL

vmihailenco:
LGTM

PR #29, opened 2016-01-15
nested_exception: fix backtrace parsing

kyrylo:
While working on https://github.com/airbrake/airbrake/pull/479 I
stumbled upon a bug in Airbrake Ruby where the default filter couldn't
process error payload because `:backtrace` was `nil`. Returning an
empty array in case the real backtrace is missing seems to be a more
secure approach.

### DIFF
--- a/lib/airbrake-ruby/nested_exception.rb
+++ b/lib/airbrake-ruby/nested_exception.rb
@@ -16,7 +16,7 @@ module Airbrake
       unwind_exceptions.map do |exception|
         { type: exception.class.name,
           message: exception.message,
-          backtrace: parse_backtrace(exception) }
+          backtrace: Backtrace.parse(exception) }
       end
     end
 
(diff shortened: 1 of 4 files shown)


kyrylo:
PTAL

lib/airbrake-ruby/nested_exception.rb
```diff
@@ -35,7 +35,7 @@ def unwind_exceptions
     end
 
     def parse_backtrace(exception)
-      return if exception.backtrace.nil? || exception.backtrace.none?
+      return [] if exception.backtrace.nil? || exception.backtrace.none?
       Backtrace.parse(exception)
```
vmihailenco: Please move this check to `Backtrace.parse` or change this method to:

```
return [] if exception.backtrace.nil? || exception.backtrace.none?
Backtrace.parse(exception.backtrace)
```

vmihailenco:
LGTM

lib/airbrake-ruby/nested_exception.rb
kyrylo: Moved the check to `Backtrace.parse`. Wanted to do the latter, but the Java exceptions check in `Backtrace` makes it a more complex refactoring (which is not worth it).

PR #33, opened 2016-01-18
rubocop: fix the Performance/TimesMap cop offenses

kyrylo:
It's a new cop introduced by
https://github.com/bbatsov/rubocop/pull/2583


vmihailenco:
LGTM

PR #34, opened 2016-01-18
rubocop: fix the Style/FrozenStringLiteralComment offenses

kyrylo:
This new cop was introduced by
https://github.com/bbatsov/rubocop/pull/2542

This feature makes strings immutable in MRI 2.3.0

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -80,3 +80,7 @@ Performance/TimesMap:
 
 Style/IfInsideElse:
   Enabled: true
+
+# TODO: enable this when Ruby 3.0 is out.
+Style/FrozenStringLiteralComment:
+  Enabled: false
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -6,23 +6,6 @@
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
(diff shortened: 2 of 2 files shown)


vmihailenco:
Build does not pass and I am not big fan of such monkey business (`# frozen_string_literal: true` at the top of every file...)

kyrylo:
I did expect a failure because we intentionally mutate strings in PayloadTruncator. I guess we could disable this cop. Does that sound good?

vmihailenco:
Sounds good, but I don't insist

kyrylo:
We should revisit this later, once immutable strings will be a common (and good practice in Ruby), which should happen in Ruby 3.0.

Disabled the cop. PTAL

.rubocop.yml
```diff
@@ -80,3 +80,6 @@ Performance/TimesMap:
 
 Style/IfInsideElse:
   Enabled: true
+
+Style/FrozenStringLiteralComment:
```
vmihailenco: Probably add a todo to revisit for Ruby 3.0 like you said previously

vmihailenco:
LGTM

.rubocop.yml
kyrylo: Added a todo. Merging.

PR #41, opened 2016-02-02
notice: attach hostname information by default

kyrylo:
The hostname information has to be present in many of our integrations
from the Airbrake gem. This information is just too useful and it
should be retreived by default.


zefer:
LGTM

vmihailenco:
:+1:

PR #45, opened 2016-02-10
Encode components in filter_url_params

jdelStrother:
I was having problems on Ruby 2.1.5 where airbrake would fail to report errors if the URL contained something like `bg_color=%23FF00FF` - filter_url_params would try to assign `url.query = "bg_color=#FF00FF"`, which raises URI::InvalidComponentError.

(Interestingly, ruby 2.2 no longer raises in this case)

WDYT?

### DIFF
--- a/lib/airbrake-ruby/filters/keys_filter.rb
+++ b/lib/airbrake-ruby/filters/keys_filter.rb
@@ -55,7 +55,13 @@ module Airbrake
 
       def filter_url_params(url)
         url.query = Hash[URI.decode_www_form(url.query)].map do |key, val|
-          should_filter?(key) ? "#{key}=[Filtered]" : "#{key}=#{val}"
+          # Ruby < 2.2 raises InvalidComponentError if the query contains
+          # invalid characters, so be sure to escape individual components.
+          if should_filter?(key)
+            "#{URI.encode_www_form_component(key)}=[Filtered]"
+          else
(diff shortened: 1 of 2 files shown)


lib/airbrake-ruby/filters/keys_filter.rb
```diff
@@ -55,7 +55,9 @@ def filter_hash(hash)
 
       def filter_url_params(url)
         url.query = Hash[URI.decode_www_form(url.query)].map do |key, val|
-          should_filter?(key) ? "#{key}=[Filtered]" : "#{key}=#{val}"
+          encoded_key = URI.encode_www_form_component(key)
+          encoded_val = URI.encode_www_form_component(val)
+          should_filter?(key) ? "#{encoded_key}=[Filtered]" : "#{encoded_key}=#{encoded_val}"
```
kyrylo: Looks like the build is failing because of line length. Could you convert this into a simple `if` statement and avoid temporary variables, so this code can be a little bit faster? Given that this code is needed to support older Rubies, could you add a brief comment describing why we need to encode components?

kyrylo:
Looks good, thanks! :) It just needs a small tweak.

jdelStrother:
Sure, updated.

kyrylo:
Fantastic, thanks! :)

PR #46, opened 2016-02-17
Support open_timeout and read_timeout

a2ikm:
Default values are Net::HTTP's.

### DIFF
--- a/lib/airbrake-ruby/config.rb
+++ b/lib/airbrake-ruby/config.rb
@@ -54,6 +54,10 @@ module Airbrake
     attr_accessor :ignore_environments
 
     ##
+    # @return [Integer] The HTTP timeout in seconds.
+    attr_accessor :timeout
+
+    ##
     # @param [Hash{Symbol=>Object}] user_config the hash to be used to build the
     #   config
     def initialize(user_config = {})
@@ -69,6 +73,8 @@ module Airbrake
       self.host = 'https://airbrake.io'
 
       self.ignore_environments = []
+
+      self.timeout = user_config[:timeout]
 
       merge(user_config)
     end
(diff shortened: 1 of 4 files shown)


kyrylo:
Thanks for the PR, but why?

a2ikm:
Two reasons:

1. Sometimes too many errors are invoked in our service. Then, we have to make sure that airbrake-gem don't wait long time.
1. We are using airbrake v4 in synchronous mode, and now moving to airbrake v5 in same mode. This is just our historical issue.

kyrylo:
What do you think about having only one option - `timeout`. It would set the value for both `read_timeout` and `open_timeout`?

kyrylo:
Gentle ping.

crunis:
My 2 cents: I think joining both options in one makes sense, it simplifies the configuration while I can't think of a situation where one would really need to specify different values in this context.
What do you think @a2ikm ?

a2ikm:
Yeah, I agree to join two options to one. It is simple and suits my purpose.

But I think it can wait for up to twice long time if setting same value to `read_timeout` and `open_timeout`.  How do you think? If it's OK, I'll fix my code.

kyrylo:
Sounds good to me.

crunis:
Increasing these values doesn't make the connection slower, it only makes slower to detect a timeout situation. If you are increasing them to cover an anomalously slow connection, in most situations you won't care what the value is as long as it covers the delay your experimenting.
So I'm also for joining them.

kyrylo:
@a2ikm let me know if you don't have time to update the PR.

a2ikm:
Sorry, I fixed the code.
Should I squash the commits?

kyrylo:
Thanks! Yes, please squash. It would also be nice to fix the Rubocop failure.

a2ikm:
I fixed Rubocop failure and squashed. Could you check that?

kyrylo:
Looks great, thanks!

a2ikm:
Thanks!

PR #47, opened 2016-02-22
Asynchronous notifier shouldn't block if queue full

crunis:
SizedQueue blocks if the queue is full. I think this behaviour is completely unexpected in an asynchronous component.

I think it would be better to drop the error in this situation. I can picture 2 scenarios where this could happen and dropping the extra errors would make sense:

1. **The app is sending errors faster than Airbrake server can process them**. It may be due to a bug in the app or because the app is being flooded by some sort of bogus request, we don't want the app to block for this reason, we would like, if possible, to keep providing service even if it's with degraded performance. In this situation, still many errors will make it to the Airbrake server so we won't miss that something is wrong.
2. **The Airbrake server has a problem and is not accepting errors**. In this situation we also don't want our app to block and stop providing service. If the Airbrake server is not working we will notice at some point, the async workers will notify the timeout, and we can fix the problem without interrupting our application's service.

I have fixed the specs accordingly

### DIFF
--- a/lib/airbrake-ruby/async_sender.rb
+++ b/lib/airbrake-ruby/async_sender.rb
@@ -24,6 +24,8 @@ module Airbrake
     #   library
     # @return [nil]
     def send(notice)
+      return will_not_deliver(notice) if @unsent.size >= @unsent.max
+
       @unsent << notice
       nil
     end
@@ -98,6 +100,18 @@ module Airbrake
           @sender.send(notice)
         end
       end
+    end
+
+    def will_not_deliver(notice)
+      backtrace = notice[:errors][0][:backtrace].map do |line|
+        "#{line[:file]}:#{line[:line]} in `#{line[:function]}'"
+      end
+      @config.logger.error(
+        "#{LOG_LABEL} AsyncSender has reached its capacity of "                   \
+        "#{@unsent.max} and the following notice will not be delivered "          \
+        "Error: #{notice[:errors][0][:type]} - #{notice[:errors][0][:message]}\n" \
+        "Backtrace: \n" + backtrace.join("\n"))
+      nil
     end
   end
 end
--- a/spec/async_sender_spec.rb
+++ b/spec/async_sender_spec.rb
@@ -6,20 +6,25 @@ RSpec.describe Airbrake::AsyncSender do
   end
 
   describe "#send" do
-    it "limits the size of the queue, but still sends all notices" do
+    it "limits the size of the queue" do
       stdout = StringIO.new
       notices_count = 1000
+      queue_size = 10
       config = Airbrake::Config.new(
-        logger: Logger.new(stdout), workers: 3, queue_size: 10
+        logger: Logger.new(stdout), workers: 3, queue_size: queue_size
       )
       sender = described_class.new(config)
       expect(sender).to have_workers
 
-      notices_count.times { |i| sender.send(i) }
+      notice = Airbrake::Notice.new(config, AirbrakeTestError.new)
+      notices_count.times { sender.send(notice) }
       sender.close
 
       log = stdout.string.split("\n")
-      expect(log.grep(/\*\*Airbrake: \{\}/).size).to eq(notices_count)
+      notices_sent    = log.grep(/\*\*Airbrake: \{\}/).size
+      notices_dropped = log.grep(/\*\*Airbrake:.*not.*delivered/).size
+      expect(notices_sent).to be >= queue_size
+      expect(notices_sent + notices_dropped).to eq(notices_count)
     end
   end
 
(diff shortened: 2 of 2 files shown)


spec/async_sender_spec.rb
```diff
@@ -54,7 +55,8 @@
 
       it "prints the debug message the correct number of times" do
         log = @stderr.string.split("\n")
-        expect(log.grep(/\*\*Airbrake: \{\}/).size).to eq(300)
+        expect(log.grep(/\*\*Airbrake: \{\}/).size).to \
+          be >= @sender.instance_variable_get(:@unsent).max
```
kyrylo: Nitpick: could you please break it into two lines like this instead:

```ruby
expect(log.grep(/\*\*Airbrake: \{\}/).size).
  to be >= @sender.instance_variable_get(:@unsent).max
```

It is more consistent with the current style.

kyrylo:
Thanks for the PR, it makes sense. Fixes https://github.com/airbrake/airbrake-ruby/issues/48.

* do you think we should log something if we drop errors?
* could you please squash?

stevecrozz:
If it helps, here's a reference to an implementation I created a while back:
https://github.com/stevecrozz/hydraulic_brake/blob/master/lib/hydraulic_brake/async_sender.rb#L31

lib/airbrake-ruby/async_sender.rb
```diff
@@ -24,7 +24,12 @@ def initialize(config)
     #   library
     # @return [nil]
     def send(notice)
-      @unsent << notice
+      if @unsent.size >= @unsent.max
+        msg = "#{LOG_LABEL} dropping notice since sending queue is full"
+        @config.logger.error msg
```
kyrylo: It was more of a question. I am not sure if we need to log, but maybe you would find it useful?
If we do want, I am not sure about the error severity. It's more like a `warn` to me, but I am open to other suggestions.

crunis: I like the idea to log, so the information it's not lost and can be retrieved if necessary.

Having the queue full of errors and starting to not send them to the error log platform it's, for me, a critical situation. If, for some reason, I'm not aware of this situation and I'm looking at the logs I would like to see it quickly since it's an ERROR. What do you think ? 

I'm almost done with the changes, I'm having some trouble with the specs but I hope to fix it soon.

crunis:
Done! I've followed @stevecrozz style.

As I was saying in the previous comment, I think is interesting to log the notice (so the information is not lost) and I would consider it with Error severity. Like I've said: "_Having the queue full of errors and starting to not send them to the error log platform it's, for me, a critical situation. If, for some reason, I'm not aware of this situation and I'm looking at the logs I would like to see it quickly since it's an ERROR_". 

What do you think ?

lib/airbrake-ruby/async_sender.rb
```diff
@@ -99,5 +101,16 @@ def spawn_worker
         end
       end
     end
+
+    def will_not_deliver(notice)
+      backtrace = notice[:errors][0][:backtrace].map do |line|
+        "#{line[:file]}:#{line[:line]} in `#{line[:function]}`"
```
kyrylo: Should be `` `#{line[:function]}' `` (no backtick in the end) if you want to make it resemble a Ruby stack frame.

lib/airbrake-ruby/async_sender.rb
```diff
@@ -24,6 +24,8 @@ def initialize(config)
     #   library
     # @return [nil]
     def send(notice)
+      return will_not_deliver(notice) if @unsent.size >= @unsent.max
```
kyrylo: Are you sure about this condition: `@unsent.size >= @unsent.max`? Shouldn't it be `@unsent.size == @unsent.max` instead? It looks like `@unsent.size` can never be greater than its capacity.

lib/airbrake-ruby/async_sender.rb
```diff
@@ -99,5 +101,16 @@ def spawn_worker
...
+    def will_not_deliver(notice)
+      backtrace = notice[:errors][0][:backtrace].map do |line|
+        "#{line[:file]}:#{line[:line]} in `#{line[:function]}`"
+      end
+      @config.logger.error(
+        "#{LOG_LABEL} AsyncSender has reached its capacity of "                   \
+        "#{@unsent.max} and the following notice will not be delivered "          \
+        "Error: #{notice[:errors][0][:type]} - #{notice[:errors][0][:message]}\n" \
+        "Backtrace: \n" + backtrace.join("\n"))
+    end
```
kyrylo: Could you please return `nil` explicitly from here, so it matches what we state on line 25?

kyrylo:
Makes sense to me, thanks for the explanation. I am happy to merge this with a couple minor fixes that I mentioned above.

lib/airbrake-ruby/async_sender.rb
crunis: Agree, but at the same time it doesn't hurt and, since there are threads around, It is more robust and cant protect against some weird race condition that could manage to increment the size value over the max. If that could happen somehow, this condition would minimise the harm (since it will still prevent from more elements being pushed into the queue).

crunis:
Applied all your fixes and commented about the condition!

kyrylo:
Looks great, thanks :+1:

crunis:
When do you plan to release the next version of airbrake-ruby @kyrylo ?

kyrylo:
I plan to release it sometime during this week. Waiting on https://github.com/airbrake/airbrake-ruby/pull/46 to be resolved and then we're ready to release :)

PR #50, opened 2016-02-26
Add option always_async

crunis:
This adds the option "always_async" that enforces that the synchronous sender is never used.

This translates into avoiding falling back to synchronous sender when all workers in the asynchronous sender are dead.

Without this option, if the Errbit server is down and dropping packets, the async workers die after the network timeout, the notifier falls back to use the synchronous sender and finally the sync sender starts blocking the application trying to deliver the errors during the timeout time.

I think it would be better that this was the default behaviour. When you read that the gem delivers errors asynchronously you don't expect a not-so-impossible scenarios (Aibrake/Errbit server down) to make the gem to start behaving synchronously, potentially blocking your application.
In this case, a 'sync_fallback' option, defaulting to false, could be added for the people that know what are they doing and want this behaviour.

But doing so should probably be considered a change of behaviour and trigger a major version increase, which seems to much for such a change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,18 @@ Airbrake.configure do |c|
 end
 ```
 
+#### always_async
+
+Always send errors asynchronously. The default value is false.
+If Airbrake.notify is called but no asynchronous workers are alive, the default behaviour is
+to send the error synchronously. Set always_async to true to prevent this behaviour.
+
+```ruby
+Airbrake.configure do |c|
+  c.always_async = true
+end
+```
+
 API
 ---
 
--- a/lib/airbrake-ruby/config.rb
+++ b/lib/airbrake-ruby/config.rb
@@ -58,6 +58,11 @@ module Airbrake
     attr_accessor :timeout
 
     ##
+    # @return [Boolean] If true, we will never fallback to synchronous delivery
+    #   of errors, as it happens by default when there are no async workers alive
+    attr_accessor :always_async
+
+    ##
     # @param [Hash{Symbol=>Object}] user_config the hash to be used to build the
     #   config
     def initialize(user_config = {})
@@ -75,6 +80,8 @@ module Airbrake
       self.ignore_environments = []
 
       self.timeout = user_config[:timeout]
+
+      self.always_async = false
 
       merge(user_config)
     end
--- a/lib/airbrake-ruby/notifier.rb
+++ b/lib/airbrake-ruby/notifier.rb
@@ -127,7 +127,7 @@ module Airbrake
     end
 
     def default_sender
-      return @async_sender if @async_sender.has_workers?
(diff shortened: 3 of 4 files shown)


vmihailenco:
>the async workers die after the network timeout

I think this should be fixed instead.

crunis:
@vmihailenco I think both things should be fixed, the async workers may die for other reasons and I don't want the airbrake-ruby gem to start blocking my applications flow.

vmihailenco:
Then we should fix other reasons too.

>and I don't want the airbrake-ruby gem to start blocking my applications flow.

Airbrake gem had only synchronous mode for years and it worked okay. So I don't believe it is a big issue.

crunis:
I guess Airbrake switched to asynchronous mode by default for a reason right?
For example, a bogus web crawler or a malicious user may hit your service with wrong URLs/parameters generating lots of errors, so it's good to know for sure that each one is not forcing your server to open a new connection to a remote server (Airbrake / Errbit).
Don't you think that's a basic measure against Denial of Service scenarios ?

kyrylo:
I agree with @vmihailenco's thinking here.

That said, the library can respawn workers, so worst case scenario is that you will be blocked for a certain period of time, but not forever. We try to keep this library as minimal as possible and adding such an option seems to be a bit too much. 

I'm not sure if our users should be concerned about these implementation details. Users care about speed, simplicity, etc. So if it delivers exceptions in a timely fashion, nobody would even notice.

The option presented in this PR might be even dangerous. What if workers can't spawn due to the lack of memory or some other unexpected scenario? I quite like the current implementation that is flexible enough.

vmihailenco:
>Don't you think that's a basic measure against Denial of Service scenarios ?

Then we should just remove sync sender, not introduce options to disable it. That said I don't believe that situation is as bad as you described it.

crunis:
I've created a PR to fix the scenario where async workers were dying: https://github.com/airbrake/airbrake-ruby/pull/52

During the week I will send a PR with a proposal to handle the situation where async workers are dead trying to spawn new ones (in a controlled fashion), instead of falling back to synchronous behaviour, sounds good ?

@kyrylo the actual behaviour is [to only spawn new async workers if a fork is detected](https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/async_sender.rb#L81), which doesn't cover scenarios with unexpected exceptions like the one I've found.

kyrylo:
@crunis my bad, sorry. You're absolutely correct. Nevertheless I am still not a fan of this option :)

crunis:
Agree that this is maybe not necessary and it's enough with something like what we are trying to do at https://github.com/airbrake/airbrake-ruby/pull/52 . Closing the PR

PR #52, opened 2016-02-29
Captures all StandardError Exceptions at sync_sender

crunis:
This exception is generated when the asynchronous worker times out trying to connect to the Airbrake server, if it's not captured, the worker dies.

### DIFF
--- a/lib/airbrake-ruby/sync_sender.rb
+++ b/lib/airbrake-ruby/sync_sender.rb
@@ -9,18 +9,6 @@ module Airbrake
     CONTENT_TYPE = 'application/json'.freeze
 
     ##
-    # @return [Array] the errors to be rescued and logged during an HTTP request
-    HTTP_ERRORS = [
-      Timeout::Error,
-      Net::HTTPBadResponse,
-      Net::HTTPHeaderSyntaxError,
-      Errno::ECONNRESET,
-      Errno::ECONNREFUSED,
-      EOFError,
-      OpenSSL::SSL::SSLError
-    ].freeze
-
-    ##
     # @param [Airbrake::Config] config
     def initialize(config)
       @config = config
@@ -39,7 +27,7 @@ module Airbrake
 
       begin
         response = https.request(req)
-      rescue *HTTP_ERRORS => ex
+      rescue => ex
         @config.logger.error("#{LOG_LABEL} HTTP error: #{ex}")
         return
       end
--- a/spec/sync_sender_spec.rb
+++ b/spec/sync_sender_spec.rb
@@ -10,4 +10,18 @@ RSpec.describe Airbrake::SyncSender do
       expect(https.read_timeout).to eq(10)
     end
   end
+
+  describe "#send" do
+    it "catches exceptions raised when sending" do
+      stdout = StringIO.new
+      config = Airbrake::Config.new(logger: Logger.new(stdout))
+      sender = described_class.new config
+      notice = Airbrake::Notice.new(config, AirbrakeTestError.new)
+      https = double("foo")
+      allow(sender).to receive(:build_https).and_return(https)
+      allow(https).to receive(:request).and_raise(StandardError.new('foo'))
+      expect(sender.send(notice)).to be_nil
+      expect(stdout.string).to match(/ERROR -- : .+ HTTP error: foo/)
+    end
+  end
 end


stevecrozz:
I'd like to ask a more generic question about the exception handling here. What is the point of rescuing only HTTP exceptions. In other words, what kind of exception can happen that you would *want* to result in a crashed async worker?

For me, I wouldn't want any kind of StandardError to result in crashed async workers. Can we change this from `rescue *HTTP_ERRORS => ex` to `rescue StandardError => ex` or simply `rescue => ex`?

stevecrozz:
Beyond the async workers though, I think its still valuable to catch all StandardError exceptions even for synchronous mode. Otherwise the caller could end having to deal with EOFError for example, or another error not in the list.

kyrylo:
Thanks for the PR and raising this question! I agree with @stevecrozz's approach here and it sounds reasonable to me. Let's just `rescue => ex` instead. If we apply such change, does it mean we don't need https://github.com/airbrake/airbrake-ruby/pull/50?

kyrylo:
Gentle ping, guys.

crunis:
Sorry @kyrylo I'm caught with something. 
Agree, this sounds enough to close https://github.com/airbrake/airbrake-ruby/pull/50 , let me just check something and come tomorrow with the PR, sounds good ?

kyrylo:
Sounds great. Don't worry and take your time.

crunis:
There you go, I've added some tests. How does it look?

crunis:
No idea why the CI specs are failing, I don't understand the error, locally the specs pass, any suggestions?

lib/airbrake-ruby/sync_sender.rb
```diff
@@ -33,18 +21,15 @@ def initialize(config)
...
-        response = https.request(req)
-      rescue *HTTP_ERRORS => ex
-        @config.logger.error("#{LOG_LABEL} HTTP error: #{ex}")
-        return
-      end
-
+      response = https.request(req)
       Response.parse(response, @config.logger)
+
+    rescue => ex
```
kyrylo: Could you only rescue `http.request`? `Response.parse` already catches exceptions and if `build_post_request` or `build_https` fail, then you can't send requests anyway.

crunis: But if `build_post_request` or `build_https` raise and exception and we don't capture it, the async worker is going to die, which is what we are trying to prevent.

kyrylo: Sure. But what's the purpose of a worker who can't work? Let it die! :)

crunis: But if all the async workers die due to a temporal error condition, we are left with no workers and then we fall back to synchronous behaviour when we don't expect it (like we were saying in https://github.com/airbrake/airbrake-ruby/pull/50)

What's the harm in capturing all exceptions in the function to avoid workers dying unnecessary and falling back to synchronous behaviour?

stevecrozz: I think its true that almost all of the exceptions will happen in https.request, but if any other RuntimeErrors occur, then a problem cascades upwards for users of the async sender. I believe it works like this:

1. If any exception here is uncaught
2. it causes the async worker thread(s) to crash.
3. and that causes AsyncSender#has_workers? to return false

Which means we fall back to the default sync sender which we're trying to avoid.

So we can rescue only https.request, but then I think we'd need another way to make sure the workers can't crash or some configuration to ensure async behavior such as #50. The intent I think is to have some way to make sure that an app (if the authors like) is only sending via background threads.

kyrylo: > @stevecrozz 
> I think its true that almost all of the exceptions will happen in https.request, but if any other RuntimeErrors occur, then a problem cascades upwards for users of the async sender.

True. However, how will we be able to fix it if we always rescue it? We're discussing this PR because something was not caught, so you had a chance to let us know that something is wrong.

If we rescue more than `http.request`, then in case of an error coming from elsewhere inside this method we won't be able to debug it. The execution will probably never reach `http.request` anyway and it presumably never fix itself. 

I am struggling to understand why we want to have a broken but alive worker. It has no use. `http.request`'s success depends on network, other nodes, weather, etc., and we can't predict everything, so it's wise to `rescue => ex` here, but the rest of that method is our code and our bugs.

Hope this makes sense. That said, perhaps I am misunderstanding something?

kyrylo: > The problem is, once the workers die, they are not resurrected. Are you open to changing that instead?

It sounds good to me in general, as a possible approach, but I reckon it's not needed right now (would be an overkill). As you said yourself "almost all of the exceptions will happen in https.request". I don't expect other code to fail, though. If it fails, then the library has a serious bug.

> the ultimate goal I have in mind isn't about keeping the workers alive, its about being able to make sure we only notify airbrake in a background thread

So if you have a broken worker (in a way as I described earlier), which you rescue on every `notify`, it would mean that you will never receive your error because it will be swallowed all the time. Currently, if workers die, you can still receive errors, which is IMO better than no errors at all.

As @vmihailenco pointed out earlier, Airbrake used to be synchronous by default for years. I understand your concerns but I believe it will be better to only rescue `https.request` (I don't expect any failures outside of that). If we will reveal that it's not enough, then we can always reconsider it. I am always open to feedback and I appreciate it a lot. So ping me any time for any [hopefully sensible, though] reason :P

crunis: Ok, I concur. Updated (and rebased) my PR, all green !

kyrylo:
Thanks! Ping me in case of trouble.

PR #54, opened 2016-03-02
filters: compare non-regexp keys exactly

kyrylo:
Users reported a bug when a string key would filter out other matching
keys. For example `Airbrake.blacklist_keys([:boom])` would also filter
out `:boombastic` key. The presented change makes sure that we compare
Strings and Symbols exactly.


zefer:
LGTM

PR #55, opened 2016-03-02
filters: filter context payload

kyrylo:
A user reported that it's not possible to filter context/user. It's a
reasonable request given that the user information has been moved to
context recently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ Airbrake Ruby Changelog
 
 ### master
 
+* Started filtering the context payload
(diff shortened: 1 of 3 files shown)


kyrylo:
This is a bit tricky. We perform URL params filtering, so with the present change it breaks (we simply replace the whole URL instead of specific params). I think stuff in the context payload should still not be filterable, that said it makes sense to filter out `user`.

kyrylo:
PTAL

zefer:
LGTM

PR #56, opened 2016-03-04
Refactor Airbrake.blacklist/whitelist_keys to be an option

### LINKED ISSUE #53: Expose blacklist_keys and whitelist_keys as config options
I just realised it makes no sense to have those as methods.

Therefore this:

```ruby
Airbrake.blacklist_keys([:user, :password, :credit_card])
```

Should become this:

```ruby
  config.blacklist_keys = [:user, :password, :credit_card]
end
```

@vmihailenco thoughts?

kyrylo:
Fixes #53 (Expose blacklist_keys and whitelist_keys as config options)

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,64 @@ Airbrake.configure do |c|
 end
 ```
 
+#### blacklist_keys
+
+Specifies which keys in the payload (parameters, session data, environment data,
+etc) should be filtered. Before sending an error, filtered keys will be
+substituted with the `[Filtered]` label.
+
+It accepts Strings, Symbols & Regexps, which represent keys of values to be
+filtered.
+
+```ruby
+Airbrake.configure do |c|
+  c.blacklist_keys = [:email, /credit/i, 'password']
+end
+
+Airbrake.notify('App crashed!', {
+  user: 'John',
+  password: 's3kr3t',
+  email: 'john@example.com',
+  credit_card: '5555555555554444'
+})
+
+# The dashboard will display this parameter as filtered, but other values won't
+# be affected:
+#   { user: 'John',
+#     password: '[Filtered]',
+#     email: '[Filtered]',
+#     credit_card: '[Filtered]' }
+```
+
+#### whitelist_keys
+
+Specifies which keys in the payload (parameters, session data, environment data,
+etc) should _not_ be filtered. All other keys will be substituted with the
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -258,6 +258,63 @@ Airbrake.configure do |c|
 end
 ```
 
+#### blacklist_keys
+
+Specifies which keys in the payload (parameters, session data, environment data,
+etc) should be filtered. Before sending an error, filtered keys will be
+substituted with the `[Filtered]` label.
+
+It accepts Strings, Symbols & Regexps, which represent keys to be filtered.
```
shifi: Nitpick: values are filtered here, so maybe change this to be more consistent with L291

lib/airbrake-ruby/config.rb
```diff
@@ -58,6 +58,18 @@ class Config
     attr_accessor :timeout
 
     ##
+    # @return [Array<String, Symbol, Regexp>] keys The keys, which should be
```
shifi: extra `keys`?

lib/airbrake-ruby/notifier.rb
```diff
@@ -60,13 +73,23 @@ def add_filter(filter = nil, &block)
...
+      )
       add_filter(Filters::KeysWhitelist.new(*keys))
     end
 
     ##
     # @macro see_public_api_method
+    # @deprecated Please use {Airbrake::Config#blacklist_keys} instead
     def blacklist_keys(keys)
+      @config.logger.warn(
+        "#{LOG_LABEL} Airbrake.whitelist_keys is deprecated. Please use the " \
```
shifi: `whitelist_keys` -> `blacklist_keys`

shifi:
Few minor comments, otherwise LGTM

README.md
kyrylo: Changed to `It accepts Strings, Symbols & Regexps, which represent keys of values to be
filtered.`, thanks!

lib/airbrake-ruby/config.rb
kyrylo: Yeah, copy-pasta :spaghetti:. Fixed.

lib/airbrake-ruby/notifier.rb
kyrylo: Fixed.

PR #57, opened 2016-03-10
notify only when there are notifiers

stevecrozz:
When there are no notifiers, this at_exit hook will cause a new exception which masks the real underlying exception:
```
lib/airbrake-ruby.rb:288:in `call_notifier': the 'default' notifier isn't configured (Airbrake::Error)
```

This makes it a bit safer, although maybe this logic should go into notify_sync instead. Any thoughts?

### DIFF
--- a/lib/airbrake-ruby.rb
+++ b/lib/airbrake-ruby.rb
@@ -296,5 +296,5 @@ end
 
 # Notify of unhandled exceptions, if there were any, but ignore SystemExit.
 at_exit do
-  Airbrake.notify_sync($ERROR_INFO) if $ERROR_INFO
+  Airbrake.notify_sync($ERROR_INFO) if $ERROR_INFO && @notifiers.any?
 end


stevecrozz:
On second thought, I think this PR is probably the way to go. If a person calls notify_sync when there are no notifiers, that does seem exception-worthy. It's only a problem in this case because it's hiding the real exception which is causing the exit.

kyrylo:
Well, I am not sure how this PR works, because `@notifiers` (at least to my expectations) should be `nil` in the `at_exit` hook.

I am aware of this inconvenience, but I couldn't come with a proper solution yet. Currently, you can still see the original error (at least in modern Rubies).

stevecrozz:
Maybe this is the wrong solution, but we do need a better solution because it is not true that you can see the original error unless you're talking about the airbrake dashboard.

In my case, I'm running an app in development environment and observing the error in my terminal with a printed backtrace (normal application development). Instead of seeing the 'real' error, I see this error about a missing defualt notifier printed to STDOUT. When I try to start my app, and my app crashes, I expect to see the real error that caused the crash.

stevecrozz:
Here's what led me to this solution:
https://github.com/airbrake/airbrake-ruby/blob/v1.1.0/lib/airbrake-ruby.rb#L281

`notify_sync` calls `call_notifier` which will raise an exception if the specified notifer doesn't exist in `@notifiers`. So if `@notifiers` really is empty or nil, then I don't see any way this could work as it exists today.

kyrylo:
> Maybe this is the wrong solution, but we do need a better solution because it is not true that you can see the original error unless you're talking about the airbrake dashboard.

We have added [support for nested exceptions](https://airbrake.io/blog/airbrake-feature/new-nested-exceptions) recently.

>  Instead of seeing the 'real' error, I see this error about a missing defualt notifier printed to STDOUT. When I try to start my app, and my app crashes, I expect to see the real error that caused the crash.

I absolutely agree with you. By the way, you must be using old Ruby if you don't see the original exception. Since Ruby 2.1+, the interpreter prints nested exceptions.

I'd love to merge your PR, but I can't because it doesn't fix the issue. I'd be very grateful if you could file an issue addressing this problem, so I could take a closer look at this. Thanks!

stevecrozz:
Actually it looks like all of this may be moot now because of #61 and I'm satisfied with that solution. Is there still a problem you're seeing that needs an issue or should we close this?

kyrylo:
Let's close for now. We can discuss later if we discover that it's really annoying :)

stevecrozz:
Ok @kyrylo, thanks.

PR #59, opened 2016-03-10
circle: update Rubinius versions and fix build

kyrylo:
The fix ensures that the build passes on newer Rubinius versions, which
fail with "NoMethodError: undefined method `join' on nil:NilClass." due
to (apparently) a race condition.

The fix is copied from https://github.com/puma/puma/pull/809.


kyrylo:
Quick fix, as without this all other PRs will be failing on Circle :(

PR #61, opened 2016-03-11
airbrake-ruby: don't report unhandled exceptions

### LINKED ISSUE #58: Configuration setting for unhandled exceptions
Hey!

Currently airbrake-ruby notifies about unhandled errors, if they cause process exit, and there is no way to prevent airbrake from doing that. What if we add a config property, that will allow to turn off notifying on such exceptions?

kyrylo:
Fixes #58 (Configuration setting for unhandled exceptions)

Although we delete this, it is still useful. Therefore, this
functionality is going to be moved to the Airbrake gem instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,10 @@ Airbrake Ruby Changelog
     c.whitelist_keys = [:page_id, 'user']
   end
   ```
-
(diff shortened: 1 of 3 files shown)


shifi:
LGTM

Issue #44, opened 2016-02-03
default notifier for multiple applications


robuye:
Hey guys,

I have 2 applications mounted in `config.ru`:

```
map '/a' do
  run ApplicationA::Application
end

map  '/b' do
  run ApplicationB::Application
end
```

`ApplicationA` is normal rails application, `ApplicationB` is a Rails engine. Both applications have Airbrake configured. It's necessary in my setup because I have separate servers running only `ApplicationA` and only `ApplicationB` too.

Configuring Airbrake standard way results in `the 'default' notifier was already configured`.
Configuring dedicated notifiers (`:application_a` and `:application_b`) results in `the 'default' notifier isn't configured`.

I understand where is this coming from, I wonder if there is a clean solution for this sort of setups. I'm fine with both applications sharing the same global config.

Thanks,

Rob

kyrylo:
Hey! Normally Rails engines are handled inside `config/routes.rb`. Are you defining that inside `config.ru` because your engine runs on a separate server? Just want to double check because that looks unconventional to me and I am not sure how you can run that on separate servers given that you define both apps in the same file. A small explanation is appreciated.

There's currently no graceful solution to configure multiple notifiers for the same integration, so you have to use the default one. That said, you can:

1. Configure the notifier only for `ApplicationA::Application` (theoretically it should just work) and `ApplicationB::Application` will simply be reusing it because our integration is a simple Rack middleware, so I assume your engine shares the same middleware stack (IIRC Rails engines itself are middlewares).
2. OR configure Airbrake in `config.ru` instead of the initializer. Simply move the code from there. Make sure that the environment is loaded so you can use `Rails.env`, `Rails.root`, etc.

robuye:
Thanks for the answer.

I have both applications mounted in `config.ru` on `Server1` and only one application mounted on `Server2`.

Server 1:
```
# config.ru
map '/a' do
  run ApplicationA::Application
end

map  '/b' do
  run ApplicationB::Application
end
```

Server 2:
```
# config.ru
map '/' do
  run ApplicationB::Application
end
```

I can't use `config/routes.rb`. There is some additional logic in place that has to happen before a request is dispatched to the right application stack. I think it's not relevant to the issue and I'd prefer not to go into implementation details here.

It's also not mutually exclusive to Rails Engines, the problem applies to any Rack application (grape, sinatra, etc.).

I appreciate the suggestions.

[...]

PR #62, opened 2016-03-12
Allow overriding airbrake configuration.

robuye:
This addresses issue #44

Given multiple application mounted in config.ru when each application
defines it's own Airbrake config server will crash with "the 'default'
notifier was already configured" error.

Example:

```
 # app1.rb
 Airbrake.configure do |c|
   #...
 end

 # app2.rb
 Airbrake.configure do |c|
   #...
 end

 # config.ru
 require 'app1'
 require 'app2' # => ERROR
```

Changes in this P/R allow the following workaround:

```
 # config.ru
 require 'airbrake'

 Airbrake.with_overrides!(notifier: :app1) do
   require 'app1'
 end

 require 'app2'
```

App1 will use :app1 notifier instead of :default now.

/cc @kyrylo

### DIFF
--- a/lib/airbrake-ruby.rb
+++ b/lib/airbrake-ruby.rb
@@ -68,6 +68,11 @@ module Airbrake
   # names, the values are Airbrake::Notifier instances.
   @notifiers = {}
 
+  ##
+  # A Hash that holds a config options. These options will override any
+  # values set via Airbrake.configure.
+  @overrides = {}
+
   class << self
     ##
     # Configures a new +notifier+ with the given name. If the name is not given,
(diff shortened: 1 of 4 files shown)


kyrylo:
Hey! Thanks for the PR. I'm afraid this won't work because the Rack integration will still be sending exceptions via the `:default` notifier. Please take a look at my PR: https://github.com/airbrake/airbrake/pull/515

I couldn't figure out how to solve the `server1.ru` issue because technically it counts as misconfiguration (if we don't take into account the existence of `server3.ru`).

kyrylo:
If we can resolve this, then I can release the new version of the library immediately. Let me know if you have some time to take a look. Thanks!

robuye:
Thanks @kyrylo I will definitely have a look over the weekend. Apologies for the delays, but I can hardly find time during the week recently.

kyrylo:
No worries! I will have some time to work on this during the weekend, no problem at all.

robuye:
Hi @kyrylo,

Sorry it's taking so long.

I'm playing with both branches on my real setup, that is main rails application, rails engine (not mounted at all, but runs as normal application on separate server) and a grape app (mounted in config.ru). My rails engine has airbrake configured in the initializer and rails runs engine initializers in a way that prevents me from wrapping the code with _overrides_ block.

I didn't have problems with grape application, since it's mounted in `config.ru` and don't share rails middleware so the `:default` middleware isn't injected via railtie.

I realise this is a bit awkward setup so if you think it's not worth support out of the box I understand it completely. I'm will try to make it work one way or another and let you know how it ends up eventually.

kyrylo:
Hey @robuye!  Thanks for letting me know. It's an interesting feature to have, so it would be nice to support it out-of-box, but I can see how tricky this is to implement. I wish I could be more useful here.

Should we close this PR, then? In the future, don't hesitate to ping me again (here, kyrylo@airbrake.io or simply with a new PR/issue).

robuye:
Yes, let's close it (and along with the associated issue), there is little value from it remaining open.

Thanks a lot for all the work you did!

PR #63, opened 2016-03-15
Don't change Net::HTTP's default of using the environment proxy

chewi:
http_proxy and https_proxy are respected by default.


kyrylo:
:+1:

PR #66, opened 2016-03-31
notifier: add the ability to append params when passing a Notice object

kyrylo:
A customer who wanted to use `notify_airbrake` from the `airbrake` gem
with own params couldn't see the params because they are never
appended. This commit fixes the bug, so you can use `notify_airbrake`
or the notify methods from this library as intended.


shifi:
LGTM

PR #70, opened 2016-04-19
filters/keys_filter: ignore bad URLs 

### LINKED ISSUE #69: send_notice raises URI::InvalidURIError with invalid URIs
If a user visits a badly formed URI (say, http://example.com/foo]bar), and the app raises an exception (perhaps because of the badly formed URI), the airbrake gem will raise a second exception while trying to parse `notice[:context][:url]`, which prevents the original error from reaching Airbrake.   Here's a partial stack from such an error:

```
/Users/jon/.rvm/rubies/ruby-2.1.5/lib/ruby/2.1.0/uri/common.rb, line 747
/Users/jon/.rvm/rubies/ruby-2.1.5/lib/ruby/2.1.0/uri/common.rb, line 1232
airbrake-ruby (1.1.0) lib/airbrake-ruby/filters/keys_filter.rb, line 32
airbrake-ruby (1.1.0) lib/airbrake-ruby/filter_chain.rb, line 64
airbrake-ruby (1.1.0) lib/airbrake-ruby/filter_chain.rb, line 62
airbrake-ruby (1.1.0) lib/airbrake-ruby/notifier.rb, line 123
airbrake-ruby (1.1.0) lib/airbrake-ruby/notifier.rb, line 45
airbrake-ruby (1.1.0) lib/airbrake-ruby.rb, line 282
airbrake-ruby (1.1.0) lib/airbrake-ruby.rb, line 136
airbrake (5.1.0) lib/airbrake/rack/middleware.rb, line 39
```

Perhaps `KeysFilter#call` should catch URI::InvalidURIError and quietly return?   Or replace `notice[:context][:url]` with `"<invalid URI>"` ?

kyrylo:
Fixes #69 (send_notice raises URI::InvalidURIError with invalid URIs)

We lose the ability to filter query params, but the good thing is that
we can deliver unmodified URL.

### DIFF
--- a/lib/airbrake-ruby/filters/keys_filter.rb
+++ b/lib/airbrake-ruby/filters/keys_filter.rb
@@ -37,10 +37,7 @@ module Airbrake
         end
 
         return unless notice[:context][:url]
-        url = URI(notice[:context][:url])
-        return if url.nil? || url.query.nil?
-
-        notice[:context][:url] = filter_url_params(url)
+        filter_url(notice)
       end
 
       ##
@@ -73,6 +70,17 @@ module Airbrake
         end.join('&')
 
         url.to_s
+      end
+
+      def filter_url(notice)
+        begin
+          url = URI(notice[:context][:url])
+        rescue URI::InvalidURIError
+          return
+        end
+
+        return unless url.query
+        notice[:context][:url] = filter_url_params(url)
       end
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Airbrake Ruby Changelog
 
 ### master
 
+* Fixed `URI::InvalidURIError` while trying to filter non-standard URLs
+  ([#70](https://github.com/airbrake/airbrake-ruby/pull/70))
+
 ### [v1.2.2][v1.2.2] (April 5, 2016)
 
 * Fixed bug in `Notifier#notify` where the `params` Hash is ignored if the first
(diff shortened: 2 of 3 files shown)


vmihailenco:
Why not to to catch URI::InvalidURIError exception and don't filter URL params?

kyrylo:
This is another option, yes. Why is it better, though?

vmihailenco:
It is better because it is the first thing that comes on mind when code throws an exception.

kyrylo:
Amended the PR to use this approach instead.

PTAL

lib/airbrake-ruby/filters/keys_filter.rb
```diff
@@ -74,6 +72,23 @@ def filter_url_params(url)
...
+
+        if url.respond_to?(:query) && url.query
+          notice[:context][:url] = filter_url_params(url)
+        end
+      end
+
+      ##
+      # Tries to convert a String to a URI, but if fails, returns the original
+      # String unchanged.
+      def parse_url(url_string)
```
vmihailenco: What is the point of the method that returns URI or string?

lib/airbrake-ruby/filters/keys_filter.rb
```diff
@@ -74,6 +72,23 @@ def filter_url_params(url)
 
         url.to_s
       end
+
+      def filter_url(notice)
```
vmihailenco: try
  url = URI(url_string)
rescue URI::InvalidURIError
  return
notice[:context][:url] = filter_url_params(url)

kyrylo:
Applied suggested changes and added one more test. 

PTAL

vmihailenco:
LGTM

PR #73, opened 2016-05-04
Correctly handle a frozen string with invalid characters

bobes:
This change fixes a `RuntimeError: can't modify frozen String` error that happens in Payload Truncator when a string with invalid characters is frozen

### DIFF
--- a/lib/airbrake-ruby/payload_truncator.rb
+++ b/lib/airbrake-ruby/payload_truncator.rb
@@ -91,7 +91,7 @@ module Airbrake
     end
 
     def truncate_string(str)
-      replace_invalid_characters!(str)
+      str = replace_invalid_characters!(str)
       return str if str.length <= @max_size
       str.slice(0, @max_size) + '[Truncated]'.freeze
     end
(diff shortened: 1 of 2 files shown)


lib/airbrake-ruby/payload_truncator.rb
```diff
@@ -109,11 +109,12 @@ def truncate_string(str)
     # @return [void]
     # @note This method mutates +str+ for speed
```
kyrylo: Could you delete this line and update the line above?

bobes: Sure. Also, would you prefer keeping the bang in the method name since it still modifies the argument when it's not frozen?

kyrylo: Good call, it makes sense. I guess it's also better to update this note, to mention that it mutates every but frozen strings.

bobes: done

lib/airbrake-ruby/payload_truncator.rb
```diff
@@ -106,14 +106,16 @@ def truncate_string(str)
     # For modern Rubies we use UTF-16 as a safe alternative.
     #
     # @param [String] str The string to replace characters
-    # @return [void]
-    # @note This method mutates +str+ for speed
+    # @return [String] a UTF-8 encoded string
+    # @note This method mutates +str+ unless it's frozen,
+    # in which case it creates a duplicate
```
kyrylo: According to YARD (the notation we use to generate docs), we should be using two spaces here:

```rb
# @note This method mutates +str+ unless it's frozen,
#   in which case it creates a duplicate
```

kyrylo:
Massive thanks for the PR, @bobes, great find & PR! 
If you could fix the nitpick (https://github.com/airbrake/airbrake-ruby/pull/73#discussion_r62020516) and squash all your commits into one, then this is ready to be merged.

bobes:
thanks for the feedback, I hope it's alright now :)

I would appreciate if you could bump the version and push it to rubygems, thank you.

kyrylo:
Sure, expect a release today. :)

kyrylo:
Just release `airbrake-ruby-1.2.4`. Happy Airbraking!

PR #32, opened 2016-01-18
airbrake-ruby: reraise SystemExit instead of raising Airbrak::Error

kyrylo:
Fixes #28 ("Airbrake::Error: the 'default' notifier isn't configured"
thrown after unsuccessful specs)

This fixes issues with the Rake integration. There was a [similar
issue][1] to a bug with the Rake integration reported a couple of weeks
ago. The change presented here fixes both issues by reraising
`SystemExit`, which propagates it and gets handled by Ruby.

[1]: https://github.com/airbrake/airbrake-ruby/issues/13

PR #75, opened 2016-05-05
airbrake-ruby: return nil instead of raising error

kyrylo:
Fixes https://github.com/airbrake/airbrake/issues/546
(Disable Airbrake when not configured?)

This is probably the most annoying thing about this library. People
often don't see the original error when we raise ours (the original
error becomes nested and it is hard to notice). Sometimes users also
want to disable Airbrake, but they can't do so because once the library
is required, it assumes that the user wants to configure it
immediately. So if you don't configure it, your app crashes.

The solution is to stop raising this error. If not the complaints, I
would probably leave it as is. The current behaviour makes a lot of
sense to me. However the amount of confusion is immense, so "we must
deal with it".

### DIFF
--- a/lib/airbrake-ruby.rb
+++ b/lib/airbrake-ruby.rb
@@ -275,20 +275,15 @@ module Airbrake
     private
 
     ##
-    # Calls +method+ on +notifier+ with provided +args+.
-    #
-    # @raise [Airbrake::Error] if none of the notifiers exist
+    # Calls +method+ on +notifier+ with provided +args+. If +notifier+ is not
+    # configured, returns nil.
     def call_notifier(notifier, method, *args, &block)
-      if @notifiers.key?(notifier)
-        @notifiers[notifier].__send__(method, *args, &block)
-      else
-        # If we raise SystemExit, the Ruby process can gracefully quit without
-        # the unwanted Airbrake::Error.
-        raise args.first if args.first.class == SystemExit
+      return unless configured?(notifier)
+      @notifiers[notifier].__send__(method, *args, &block)
+    end
 
-        raise Airbrake::Error,
-              "the '#{notifier}' notifier isn't configured"
-      end
+    def configured?(notifier)
+      @notifiers.key?(notifier)
     end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ Airbrake Ruby Changelog
 
 ### master
 
+* **IMPORTANT:** stopped raising the `the 'default' notifier isn't configured`
+  error when Airbrake is not configured
+  ([#75](https://github.com/airbrake/airbrake-ruby/pull/75))
+
 ### [v1.2.4][v1.2.4] (May 4, 2016)
 
 * Fixed bug when trying to truncate frozen strings
--- a/spec/airbrake_spec.rb
+++ b/spec/airbrake_spec.rb
@@ -23,10 +23,7 @@ RSpec.describe Airbrake do
   shared_examples 'error handling' do |method|
     it "raises error if there is no notifier when using #{method}" do
       described_class.instance_variable_set(:@notifiers, {})
-
-      expect { described_class.__send__(method, 'bingo') }.
-        to raise_error(Airbrake::Error,
-                       "the 'default' notifier isn't configured")
+      expect(described_class.__send__(method, 'bingo')).to be_nil
     end
   end
 


lib/airbrake-ruby.rb
```diff
@@ -275,20 +275,16 @@ def create_deploy(deploy_params, notifier = :default)
...
-        @notifiers[notifier].__send__(method, *args, &block)
-      else
-        # If we raise SystemExit, the Ruby process can gracefully quit without
-        # the unwanted Airbrake::Error.
-        raise args.first if args.first.class == SystemExit
-
-        raise Airbrake::Error,
-              "the '#{notifier}' notifier isn't configured"
+      unless @notifiers.key?(notifier)
+        return { 'error' => "the '#{notifier}' notifier isn't configured" }
```
zefer: Could it be helpful to explain the consequences in this message, to help explain what this means and to help decide if they want to fix it? Also, is it obvious to the user that this is related to Airbrake? Maybe something like: `{ 'error' => "the '#{notifier}' notifier isn't configured, Airbrake will not be notified of errors" }`

kyrylo: > Could it be helpful to explain the consequences in this message, to help explain what this means and to help decide if they want to fix it?

You mean, in the readme?

> Also, is it obvious to the user that this is related to Airbrake

I guess it can be more obvious, yes.

zefer: > You mean, in the readme?

No, in the message. I tried to resolve both points I made with my suggestion.

kyrylo: How about: `**Airbrake: the '#{notifier}' notifier isn't configured, '#{method}' was not called.`.

This error can occur even if you call `add_filter`, for example, so if you call it and see `Airbrake will not be notified of errors`, it might be a little bit confusing. On the other hand, I doubt anybody will be checking the return value of `add_filter`. Thoughts?

zefer: 👍 I like it!

kyrylo: Amended the message.

kyrylo:
PTAL

lib/airbrake-ruby.rb
```diff
@@ -275,20 +275,19 @@ def create_deploy(deploy_params, notifier = :default)
...
-        # the unwanted Airbrake::Error.
-        raise args.first if args.first.class == SystemExit
-
-        raise Airbrake::Error,
-              "the '#{notifier}' notifier isn't configured"
+      unless @notifiers.key?(notifier)
+        return {
+          'error' => "#{LOG_LABEL} the '#{notifier}' notifier isn't configured, " \
+                     "'#{method}' was not called"
+        }
```
vmihailenco: Why not to log the same error message? What is the point of returning this if none really will handle it?

lib/airbrake-ruby.rb
```diff
@@ -275,20 +275,19 @@ def create_deploy(deploy_params, notifier = :default)
...
     #
-    # @raise [Airbrake::Error] if none of the notifiers exist
+    # @return [Hash{String=>String}, Object]
     def call_notifier(notifier, method, *args, &block)
-      if @notifiers.key?(notifier)
-        @notifiers[notifier].__send__(method, *args, &block)
-      else
-        # If we raise SystemExit, the Ruby process can gracefully quit without
-        # the unwanted Airbrake::Error.
-        raise args.first if args.first.class == SystemExit
```
vmihailenco: Why not to keep this?

lib/airbrake-ruby.rb
```diff
@@ -275,20 +275,19 @@ def create_deploy(deploy_params, notifier = :default)
...
-
-        raise Airbrake::Error,
-              "the '#{notifier}' notifier isn't configured"
+      unless @notifiers.key?(notifier)
```
vmihailenco: This could read better if it is `unless @configured`

lib/airbrake-ruby.rb
kyrylo: I was thinking about doing it like so, but the problem is that we log through user configured loggers, via `@config.logger.debug(...)`. If none of the notifiers exist, it also means loggers don't exist. I thought about logging to STDERR, but decided not to do so because there's a risk of polluting it with countless messages, which is something that people who want to disable Airbrake would hate.

lib/airbrake-ruby.rb
kyrylo: This is needed only if we raise `Airbrake::Error`. Context: https://github.com/airbrake/airbrake-ruby/pull/32

lib/airbrake-ruby.rb
kyrylo: `@configured` would make sense if we were in the Notifier class, but we're inside an Airbrake "module". Or do you mean something else?

vmihailenco: why it can't be

```
alias configured = @notifiers.key?(notifier)
```

lib/airbrake-ruby.rb
vmihailenco: The just remove this return. None will ever use it.

kyrylo: 1. We still need to return early in order to prevent execution (which will crash), so I assume you want to remove the hash
2. I can see your point and I partially agree with it, but my question is what harm will this hash introduce? If nobody uses the return value, it's still useful. If a customer complains that Airbrake doesn't send errors, my first question will be "what is the return value?". Think of it as a support tool.

lib/airbrake-ruby.rb
kyrylo: Just pushed something similar https://github.com/airbrake/airbrake-ruby/pull/75/commits/43bf37d1f2bca239977a3fe83f7eb9fbaeb23a20

PTAL

lib/airbrake-ruby.rb
vmihailenco: This looks wrong to me, because now every notifier method can return hash `{error: "qqqq"}` (even method that return bool etc) and that is a bad design.

lib/airbrake-ruby.rb
vmihailenco: that is it, thanks

lib/airbrake-ruby.rb
kyrylo: Stopped returning the Hash. PTAL

vmihailenco:
LGTM, but I would add a log message

kyrylo:
Added a log message.

khustochka:
After trying this branch I am getting when Airbrake is not configured:

undefined method `[]' for nil:NilClass

airbrake (5.2.3) lib/airbrake/rack/notice_builder.rb:49:in `block in <class:NoticeBuilder>'
airbrake (5.2.3) lib/airbrake/rack/notice_builder.rb:43:in `block in build_notice'
airbrake (5.2.3) lib/airbrake/rack/notice_builder.rb:43:in `each'
airbrake (5.2.3) lib/airbrake/rack/notice_builder.rb:43:in `build_notice'
airbrake (5.2.3) lib/airbrake/rack/middleware.rb:38:in `notify_airbrake'
airbrake (5.2.3) lib/airbrake/rack/middleware.rb:24:in `rescue in call'
airbrake (5.2.3) lib/airbrake/rack/middleware.rb:21:in `call'

kyrylo:
@khustochka thanks for letting me know! I'll prepare a fix tomorrow.

PR #76, opened 2016-05-05
Release v1.2.5

kyrylo:


kyrylo:
I'd like to postpone this as I have more changes that need to be included.

kyrylo:
This should be a MINOR version bump because the public API introduced some breaking changes. Closed in favour of https://github.com/airbrake/airbrake-ruby/pull/78.

PR #77, opened 2016-05-09
README: fix asynchronous examples

kyrylo:
A user reported that they were confused our examples don't work. Same
old problem here: the program finishes before our workers can finish
work, so we must wait for them.


vmihailenco:
LGTM

Issue #80, opened 2016-05-11
Airbrake: unexpected code (405). Body: Method Not Allowed


newdark:
This error above was caused by something that took me a while to figure out. I am running airbrake (5.0.5) on my application and found that this error comes up when you are setting the `project_id` and the `project_key` to blank values. 

`config/initializers/airbrake.rb`
```ruby
Airbrake.configure do |c|
  c.project_id = ''
  c.project_key = ''
```
```shell
rake airbrake:deploy USERNAME=my_user ENVIRONMENT=staging REVISION=revision REPOSITORY=github_repo VERSION=1
````
results from command

```shell
Request URI: /api/v4/projects/deploys?key=
Req: #<Net::HTTP::Post POST>
Endpoint: #<URI::HTTPS https://airbrake.io/api/v4/projects/deploys?key=>
Notice: {:environment=>"alpha", :username=>"my_user", :revision=>"revision", :repository=>"github_repo", :version=>"1"}
Response: #<Net::HTTPMethodNotAllowed 405 Method Not Allowed readbody=true>
E, [2016-05-11T11:55:50.492599 #97422] ERROR -- : **Airbrake: unexpected code (405). Body: Method Not Allowed
```

## My Understanding of Results

Most of what I was seeing now makes sense. based off what I see above I get why it resulted in the 405. Because the **project_id** is not present it changes the URL. So a 405 is exactly what you expect. when you don't have a **project_id**

```
Request URI: /api/v4/projects/{project_id}/deploys?key={project_key}
Req: #<Net::HTTP::Post POST>
Endpoint: #<URI::HTTPS https://airbrake.io/api/v4/projects/{project_id}/deploys?key=39f22e0d8d5f97df880bd62567b0f2b6>
Notice: {:environment=>"alpha", :username=>"my_user", :revision=>"d2eae2dd353f67f7c95f97b7fcbe23ff170382c6", :repository=>"github_repo", :version=>"1"}
Response: #<Net::HTTPCreated 201 Created readbody=true>
```

### Possible Solution

I think a good idea would be to add a check to make sure that the project_id and project key are present before you send the post. If not present warn the user but perform the action anyway. As you could be in the test environment or in the development environment. However there is much more to take into account as you can run the rake airbrake:deploy using RAILS_ENV=development and ENVIRONMENT=staging. So might be be also a solution to only warn if the ENVIRONMENT that you are deploying to is not in the ignore_environemnts varable. 

[...]

PR #81, opened 2016-05-11
#80 | Warn users that the project_key and project_id are blank

newdark:


kyrylo:
Thanks for the PR. I think at the moment the best way to handle this would be checking the emptiness [here](https://github.com/airbrake/airbrake-ruby/blob/80bf59146a0b20856ef359908a9cd963d29fffef/lib/airbrake-ruby/notifier.rb#L33), so we could raise the same exception. Could you refactor your approach?

newdark:
@kyrylo So the reason I was adding a warning and not a raise error is because in testing and development mode we don't care about those keys. I found that because of this I would work around the problem by calling to_s on the object so the system will not raise _both :project_id and :project_key are required_. It seams silly but not having to provide a project_id or project_key is a valid use case because there are some rails envs that don't require it. If you know a way to have it ignore the test and development environment when so it does not raise I think that might be the solution.

kyrylo:
I'm aware that people find the raise behaviour a bit confusing and I'll probably change the way it works. However, I also think that this PR should not introduce diversion and work the same way as the check I linked to. By the way, to work around it, you could set `project_{id,key}` to a dummy value. 

In short, it's better to be consistent and a bit misleading, than inconsistent and misleading. Hope this makes sense.

newdark:
I think you're right, this code was more as an example of a solution rather than actually solution to the problem. I am going to close this now and hopefully this will allow a better solution to be implemented. As I have no idea on what will work here or what the impact of changes could have on the tool. Anyway hope this helps.

kyrylo:
Sounds good, thanks. Let's keep https://github.com/airbrake/airbrake-ruby/issues/80 open for now.

PR #85, opened 2016-05-25
notifier: don't try to construct a notice if the current env is ignored

kyrylo:
Fixes https://github.com/airbrake/airbrake/issues/554 (Pipeline + React
stopping Airbrake with non-descript Error instead of Explanitory Rails
Error.)

We want to return as early as possible, so we don't try to parse the
given exception, which ensures we never have to deal with bad
backtraces.


vmihailenco:
LGTM

PR #87, opened 2016-06-03
config: don't raise when current environment is ignored

### LINKED ISSUE #80: Airbrake: unexpected code (405). Body: Method Not Allowed

kyrylo:
Fixes #80 (Airbrake: unexpected code (405). Body: Method Not Allowed)
and many other complains of people who are confused by the current
behaviour when we raise an error when they ignore current environment.

### DIFF
--- a/lib/airbrake-ruby/config.rb
+++ b/lib/airbrake-ruby/config.rb
@@ -125,6 +125,22 @@ module Airbrake
       self
     end
 
+    ##
+    # @return [Boolean] true if the config meets the requirements, false
+    #   otherwise
+    def valid?
+      return true if ignored_environment?
+      return false unless project_id.is_a?(Integer)
+      project_key.is_a?(String) && !project_key.empty?
+    end
+
+    ##
+    # @return [Boolean] true if the config ignores current environment, false
+    #   otherwise
+    def ignored_environment?
+      ignore_environments.include?(environment)
+    end
+
     private
 
     def set_option(option, value)
--- a/lib/airbrake-ruby/notifier.rb
+++ b/lib/airbrake-ruby/notifier.rb
@@ -30,7 +30,7 @@ module Airbrake
     def initialize(user_config)
       @config = (user_config.is_a?(Config) ? user_config : Config.new(user_config))
 
-      unless [@config.project_id, @config.project_key].all?
+      unless @config.valid?
         raise Airbrake::Error, 'both :project_id and :project_key are required'
       end
 
@@ -143,9 +143,7 @@ module Airbrake
     end
 
     def send_notice(exception, params, sender = default_sender)
-      if @config.ignore_environments.any?
-        return if @config.ignore_environments.include?(@config.environment)
-      end
+      return if @config.ignored_environment?
 
       notice = build_notice(exception, params)
       @filter_chain.refine(notice)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ Airbrake Ruby Changelog
 
 ### master
 
+* Stopped raising error when the notifier lacks either project ID or project key
+  and also told to ignore current environment. As the result, empty string for
+  `project_key` is also validated now (forbidden)
+  ([#87](https://github.com/airbrake/airbrake-ruby/pull/87))
+
 ### [v1.3.2][v1.3.2] (May 27, 2016)
 
 * Fixed bug when the library raises unwanted exception, when current environment
--- a/spec/config_spec.rb
+++ b/spec/config_spec.rb
@@ -76,4 +76,75 @@ RSpec.describe Airbrake::Config do
       end
     end
   end
+
+  describe "#valid?" do
+    context "when project_id is nil" do
+      it "returns false" do
+        config.project_id = nil
+        config.project_key = '123'
+
+        expect(config).not_to be_valid
+      end
+    end
+
+    context "when project_key is nil" do
+      it "returns false" do
+        config.project_id = 123
+        config.project_key = nil
+
+        expect(config).not_to be_valid
+      end
+    end
+
+    context "when the current environment is ignored" do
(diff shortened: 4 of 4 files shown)


lib/airbrake-ruby/config.rb
```diff
@@ -125,6 +125,23 @@ def merge(config_hash)
...
+      return true if ignored_environment?
+      return false unless project_id.is_a?(Integer)
+      project_key.is_a?(String) && !project_key.empty?
+    end
+
+    ##
+    # @return [Boolean] true if the config ignores current environment, false
+    #   otherwise
+    def ignored_environment?
+      return false if ignore_environments.none?
```
sgray11: This seems like an unnecessary early return. The next line should cover everything, right?

kyrylo: Correct, good catch, fixed.

sgray11:
LGTM pending CI

vmihailenco:
LGTM

lib/airbrake-ruby/notifier.rb
```diff
@@ -30,7 +30,7 @@ class Notifier
     def initialize(user_config)
       @config = (user_config.is_a?(Config) ? user_config : Config.new(user_config))
 
-      unless [@config.project_id, @config.project_key].all?
```
mvbehr: @kyrylo 
This change got me into a bit of a trouble.
I upgraded from `v1.3.1` to `v1.4.1` and it started to complain about `'both :project_id and :project_key are required'`
But I had these variables set in my `.env` file.

After re-checking the changelog I found out, that I had my `project_id` set to `"test"` in development environment.
And now it needs to be an integer.
Maybe you can change the error message to be more explicit?
`':project_id and :project_key are required as integer'`

kyrylo: Try a newer version (1.4.2). It was fixed by https://github.com/airbrake/airbrake-ruby/pull/89 :)
Sorry for the inconvenience.

mvbehr: No worries! :)
I am referring to the error message itself. 
It tells me, that the two variables are required, when both values were already set.
So I had to look at the code to get to know it has to be an Integer.

Anyway, I am fine with the state right now and also think its just an edge case.
Thanks!

PR #89, opened 2016-06-06
config: allow setting String numbers as project_id

kyrylo:
It was a bad idea to restrict this option to Integers. People often
configure this option via ENV variables, so the value would be a
String.

### DIFF
--- a/spec/config_spec.rb
+++ b/spec/config_spec.rb
@@ -120,12 +120,21 @@ RSpec.describe Airbrake::Config do
       end
     end
 
-    context "when the project_id value is not an Integer" do
+    context "when the project_id value is not an number" do
+      it "returns false" do
+        config.project_id = 'bingo'
+        config.project_key = '321'
+
+        expect(config).not_to be_valid
+      end
+    end
+
+    context "when the project_id value is a String number" do
       it "returns false" do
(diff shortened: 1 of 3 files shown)


spec/config_spec.rb
```diff
@@ -120,15 +120,24 @@
       end
     end
 
-    context "when the project_id value is not an Integer" do
+    context "when the project_id value is not an number" do
```
sgray11: Nitpick: `a number`

spec/config_spec.rb
```diff
@@ -120,15 +120,24 @@
...
-        config.project_id = '123'
+        config.project_id = 'bingo'
         config.project_key = '321'
 
         expect(config).not_to be_valid
       end
     end
 
+    context "when the project_id value is a String number" do
+      it "returns false" do
```
sgray11: `returns true`

spec/config_spec.rb
kyrylo: Fixed.

spec/config_spec.rb
kyrylo: Fixed.

sgray11:
LGTM!

kyrylo:
👍

PR #92, opened 2016-06-08
config: warn if ignore_environments is set, but environment isn't

### LINKED ISSUE #91: Airbrake being run on test and development environments
Hey, 

My team is experiencing a problem with the new version of Airbrake. We are receiving error messages from our development and test environments even though we have
 config.ignore_environments = %w(development test)
in our config block.

Please advise what we may have done wrong :).

Richard

kyrylo:
Fixes #91 (Airbrake being run on test and development environments)

This is not the first time someone is confused about this behaviour.

### DIFF
--- a/lib/airbrake-ruby/config.rb
+++ b/lib/airbrake-ruby/config.rb
@@ -138,6 +138,11 @@ module Airbrake
     # @return [Boolean] true if the config ignores current environment, false
     #   otherwise
     def ignored_environment?
+      if ignore_environments.any? && environment.nil?
+        logger.warn("#{LOG_LABEL} the 'environment' option is not set, " \
+                    "'ignore_environments' has no effect")
+      end
+
       ignore_environments.include?(environment)
     end
 
--- a/spec/config_spec.rb
+++ b/spec/config_spec.rb
@@ -156,4 +156,37 @@ RSpec.describe Airbrake::Config do
       end
     end
   end
+
+  describe "#ignored_environment?" do
+    describe "warnings" do
+      let(:out) { StringIO.new }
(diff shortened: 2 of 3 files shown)


lib/airbrake-ruby/config.rb
```diff
@@ -138,6 +138,11 @@ def valid?
     # @return [Boolean] true if the config ignores current environment, false
     #   otherwise
     def ignored_environment?
+      if ignore_environments.any? && environment.nil?
+        logger.warn("#{LOG_LABEL} the 'environment' option is not set, " \
+                    "'ignore_environments' has no effect")
```
shifi: Can you add test for this scenario? Thanks.

kyrylo: Sure, added tests.

kyrylo:
PTAL

spec/config_spec.rb
```diff
@@ -156,4 +156,41 @@
       end
     end
   end
+
+  describe "#ignored_environment?" do
+    describe "warnings" do
+      context "when 'ignore_environments' is set and 'environment' isn't" do
+        it "prints a warning" do
+          out = StringIO.new
+          config = described_class.new(logger: Logger.new(out))
```
shifi: This two lines are the same for all tests here:
```ruby
out = StringIO.new
config = described_class.new(logger: Logger.new(out))
```
Should you move them to `let` block inside `describe "warning" do`?

shifi:
LGTM

spec/config_spec.rb
kyrylo: That would work. Moved, thanks.

PR #94, opened 2016-06-10
config: do not rely on types when trying to compare environments

kyrylo:
Fixes https://github.com/airbrake/airbrake/issues/558 (Airbrake
reports during rspec)

This is also a tiny wart, which confused many people. Relaxing this
restriction doesn't harm anyone.


sgray11:
LGTM

PR #99, opened 2016-07-08
backtrace: add support for PL/SQL exceptions

### LINKED ISSUE #98: Airbrake throws away information for incompatible backtraces
Hi,

This pull request (from what i gather) introduces a new method of parsing backtraces:

https://github.com/airbrake/airbrake-ruby/pull/4

However it errors out early when it fails to parse, throwing away the rest of the backtrace:

    raise Airbrake::Error, "can't parse '#{stackframe}'"

In my case an Oracle stored procedure call might fail, their errors are incompatible with your parser, but instead of printing the whole thing, only the last item is returned:

    Airbrake::Error (can't parse 'stmt.c:243:in oci8lib_220.so')

Example of the full "backtrace"/error message:

    OCIError: ORA-02291: integrity constraint (STORE.SN_PACE_REGISTRATIONS_FK01) violated - parent key not found
    ORA-06512: at "STORE.LI_LICENSES_PACK", line 1945
    ORA-06512: at "ACTIVATION.LI_ACT_LICENSES_PACK", line 101
    ORA-06512: at line 2
    from stmt.c:243:in oci8lib_220.bundle

Downgrading to 4.3.8 resolves the problem for me giving me:

    ORA-02291: integrity constraint (STORE.SN_PACE_REGISTRATIONS_FK01) violated - parent key not found ORA-06512: at "STORE.LI_LICENSES_PACK", line 1945 ORA-06512: at "ACTIVATION.LI_ACT_LICENSES_PACK", line 101 ORA-06512: at line 2

Note: This not only breaks airbrake integration, but hides error messages in both development mode webview and log files.

* Airbrake version: 5.4.0
* Ruby version: ruby 2.2.4p230 (2015-12-16 revision 53155) [x86_64-darwin15]
* Framework name & version: rails (4.2.6), ruby-oci8 (2.2.2), ruby-plsql (0.6.0)

kyrylo:
Fixes #98
(Airbrake throws away information for incompatible backtraces)

### DIFF
--- a/lib/airbrake-ruby/backtrace.rb
+++ b/lib/airbrake-ruby/backtrace.rb
@@ -38,11 +38,31 @@ module Airbrake
     # @return [Regexp] the template that tries to assume what a generic stack
     #   frame might look like, when exception's backtrace is set manually.
     GENERIC_STACKFRAME_REGEXP = %r{\A
+      (?:from\s)?
       (?<file>.+)              # Matches '/foo/bar/baz.ext'
       :
       (?<line>\d+)?            # Matches '43' or nothing
-      (in\s`(?<function>.+)')? # Matches "in `func'" or nothing
+      (?:
+        in\s`(?<function>.+)'  # Matches "in `func'"
+      |
+        :in\s(?<function>.+)   # Matches ":in func"
+      )?                       # ... or nothing
     \z}x
+
+    ##
+    # @return [Regexp] the template that matches exceptions from PL/SQL such as
+    #   ORA-06512: at "STORE.LI_LICENSES_PACK", line 1945
+    # @note This is raised by https://github.com/kubo/ruby-oci8
+    OCI_STACKFRAME_REGEXP = /\A
+      (?:
+        ORA-\d{5}
+        :\sat\s
+        (?:"(?<function>.+)",\s)?
+        line\s(?<line>\d+)
+      |
+        #{GENERIC_STACKFRAME_REGEXP}
+      )
+    \z/x
 
     ##
     # Parses an exception's backtrace.
@@ -55,6 +75,8 @@ module Airbrake
 
       regexp = if java_exception?(exception)
                  JAVA_STACKFRAME_REGEXP
+               elsif oci_exception?(exception)
+                 OCI_STACKFRAME_REGEXP
                else
                  RUBY_STACKFRAME_REGEXP
                end
@@ -76,6 +98,10 @@ module Airbrake
 
     class << self
       private
+
+      def oci_exception?(exception)
+        defined?(OCIError) && exception.is_a?(OCIError)
+      end
 
       def stack_frame(match)
         { file: match[:file],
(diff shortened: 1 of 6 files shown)


spec/backtrace_spec.rb
```diff
@@ -146,5 +146,32 @@
...
+
+      let(:ex) do
+        OCIError = AirbrakeTestError
+        OCIError.new.tap { |e| e.set_backtrace(bt) }
+      end
+
+      after { Object.__send__(:remove_const, :OCIError) }
+
+      let(:parsed_backtrace) do
+        [{ file: 'ORA-06512', line: 1945, function: 'STORE.LI_LICENSES_PACK' },
```
vmihailenco: `ORA-06512` does not look like a file. I think we can omit it...

vmihailenco: My guess is that ORA-06512 should go to error message or error type.

lib/airbrake-ruby/backtrace.rb
```diff
@@ -38,13 +38,32 @@ module Backtrace
...
     GENERIC_STACKFRAME_REGEXP = %r{\A
+      (?:from\s)?
       (?<file>.+)              # Matches '/foo/bar/baz.ext'
       :
       (?<line>\d+)?            # Matches '43' or nothing
-      (in\s`(?<function>.+)')? # Matches "in `func'" or nothing
+      (?:
+        in\s`(?<function>.+)'    # Matches "in `func'"
+      |
+        :in\s(?<function>.+\..+) # Matches ":in func"
```
vmihailenco: It is not clear why you require dot `\.` in function name...

kyrylo: Good call, fixed, so it doesn't require it.

spec/backtrace_spec.rb
kyrylo: I guess our best bet is to simply remove it, then.

Here's an error and type from such exception:
```
OCIError: ORA-02291: integrity constraint (STORE.SN_PACE_REGISTRATIONS_FK01) violated - parent key not found
```

OCIError is the type and the rest is the message.

kyrylo:
I added a second commit, which removes raising. We now log the error and still parse the backtrace (see tests).

PTAL

spec/backtrace_spec.rb
```diff
@@ -143,7 +160,38 @@
...
+         %(ORA-06512: at line 2),
+         %(from stmt.c:243:in oci8lib_220.bundle)]
+      end
+
+      let(:ex) do
+        OCIError = AirbrakeTestError
+        OCIError.new.tap { |e| e.set_backtrace(bt) }
+      end
+
+      after { Object.__send__(:remove_const, :OCIError) }
```
vmihailenco: `OCIError = AirbrakeTestError` and this line look strange and probably deserve a comment.

vmihailenco:
LGTM

spec/backtrace_spec.rb
kyrylo: Actually, I realised it can be replaced with `stub_const` from RSpec.

PR #107, opened 2016-08-11
backtrace: add support for ExecJS backtraces

### LINKED ISSUE #103: "can't parse '#{stackframe}' (please file an issue so we can fix " \
I got this particular error which led me here, I am filing an issue because the code asked to!

I ended up finding the error that broke airbrake, working in CoffeeScript with better errors that error kept popping up. 

I found the problem, my JS tendencies had popped up and I had assigned `var x = []` instead of `x = []`, causing this error to pop up. 

Thanks!

kyrylo:
Fixes #103 ("can't parse '#{stackframe}' (please file an issue so we
 can fix " \")")

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Airbrake Ruby Changelog
 
 ### master
 
+* Added support for CoffeeScript/ExecJS backtraces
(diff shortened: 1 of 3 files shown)


vmihailenco:
LGTM

PR #110, opened 2016-08-17
backtrace: fix ExecJS backtraces support for Ruby 1.9.3

kyrylo:
A user reported that they see a `NameError` with a Rails 3 and Ruby 1.9
app. The problem is that `Exception#cause` is not supported on
Ruby 1.9, so it needs special handling.

Also fixes Rubocop error about the Backtrace module being too long (refactored).

### DIFF
--- a/spec/backtrace_spec.rb
+++ b/spec/backtrace_spec.rb
@@ -191,7 +191,7 @@ RSpec.describe Airbrake::Backtrace do
       end
     end
 
-    context "given an ExecJS backtrace" do
+    context "given an ExecJS exception" do
       let(:bt) do
         ['compile ((execjs):6692:19)',
          'eval (<anonymous>:1:10)',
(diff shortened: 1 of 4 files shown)


spec/backtrace_spec.rb
```diff
@@ -218,11 +216,35 @@
...
+
+        it "returns a properly formatted array of hashes" do
+          stub_const('ExecJS::RuntimeError', AirbrakeTestError)
+          stub_const('Airbrake::RUBY_19', false)
+
+          expect(
+            described_class.parse(ex, Logger.new('/dev/null'))
+          ).to eq(parsed_backtrace)
+        end
+      end
```
vmihailenco: Why we simply don't run tests on Ruby 1.9?

vmihailenco:
LGTM

spec/backtrace_spec.rb
kyrylo: We do :) We didn't have a test for the failing scenario, so we never reached the failing check in our existing test.

PR #113, opened 2016-08-26
Add support for custom exception attributes

### LINKED ISSUE #112: Adding to the 'params' area of airbrake.io when Airbrake.notify not explicitly called
airbrake (5.3.0), airbrake-ruby (1.4.2), ruby 2.2.3p173

Passing custom params is straightforward when explicitly notifying Airbrake:`Airbrake.notify("Temp is too high", params: { max_temp: 300 }`

How can extra params be passed when Airbrake is notified implicitly e.g. from a failed request or failed job in Sidekiq / ActiveJob?

---

I gather from https://github.com/airbrake/airbrake/issues/528 that this might be possible using the `add_filter` API but I don't think `notice` exposes the exception object itself.

(Looking here - https://github.com/airbrake/airbrake-ruby/blob/e426a94fd90080d76b04ddecaa2448353a2f58fc/lib/airbrake-ruby/nested_exception.rb#L16 - we only get the class name, message and backtrace of exceptions. No way to access arbitrary attributes on the exception.)

e.g. 

```ruby
  class Error < StandardError
    attr_reader :params

    def initialize(params = {})
      @params = params
    end
  end

  # if it were possible to expose the exception object on `notice`:
  class AddCustomParamsFilter
    def call(notice)
      notice[:errors].each do |error|
        notice[:params].merge!(error[:type] => error[:exception].params)
      end
    end
  end
```

------

Stepping back a bit - what I'd like to do is define exception classes with custom attributes and have those attributes pass over to Airbrake. At the moment I'm including all the required information in the `#message`, which is messy and prevents the UI from deduping the errors properly.

kyrylo:
Fixes #112 (Adding to the 'params' area of airbrake.io when
Airbrake.notify not explicitly called)

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ Key features
 * Support for Java exceptions occurring in JRuby
 * SSL support (all communication with Airbrake is encrypted by default)
 * Support for fatal exceptions (the ones that terminate your program)
+* Support for custom exception attributes<sup>[[link](#custom-exception-attributes)]</sup>
 * Last but not least, we follow semantic versioning 2.0.0<sup>[[link][semver2]]</sup>
 
 Installation
@@ -527,6 +528,38 @@ filtered. Modifiable payload includes:
 ```ruby
 notice[:params][:my_param] = 'foobar'
 ```
+
+### Custom exception attributes
+
+The library supports custom exception attributes. This is useful if you work
+with custom exceptions, which define non-standard attributes and you can't
+attach any additional data with help of the [`add_filter`](#airbrakeadd_filter)
+API due to the fact that the data isn't available at confuration time yet.
+
+In this case, you could define a special hook method on your exception called
+`#to_airbrake`. The method must return a Hash the keys of which must be a subset
+of the ones mentioned in the [`Notice#[]`](#notice--notice) API.
+
+```ruby
+class MyException
+  def initialize
+    @http_code = 404
+  end
+
+  # The library expects you to define this method. You must return a Hash,
+  # containing the keys you want to modify.
+  def to_airbrake
+    { params: { http_code: @http_code } }
+  end
+end
+
+# The `{ http_code: 404 }` Hash will transported to the Airbrake dashboard via
+# the `#to_airbrake` method.
+Airbrake.notify(MyException.new)
+```
+
+Note: you don't have to call `Airbrake.notify` manually to be able to benefit
+from this API. It should "just work".
 
 Additional notes
 ----------------
--- a/lib/airbrake-ruby/notice.rb
+++ b/lib/airbrake-ruby/notice.rb
@@ -2,7 +2,9 @@ module Airbrake
   ##
   # Represents a chunk of information that is meant to be either sent to
   # Airbrake or ignored completely.
+  # rubocop:disable Metrics/ClassLength
   class Notice
+    ##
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -528,6 +529,38 @@ filtered. Modifiable payload includes:
...
+
+```ruby
+class MyException
+  def initialize
+    @http_code = 404
+  end
+
+  # The library expects you to define this method. You must return a Hash,
+  # containing the keys you want to modify.
+  def to_airbrake
```
vmihailenco: Why not to use some standard Ruby method like `to_hash` and/or `to_json`?

kyrylo: When somebody defines a method with this name it's already clear why. It also must correspond to our API scheme, so if this was `to_hash`, I would expect it to return this instead: `{ http_code: 404 }`. But we need the `params` key, for example.

vmihailenco: I am not convinced. You use `to_hash` or something like that to serialize arrays, hashes and classes. You can just set `notices.params[exception] = exc` and let serializer do its job.

>But we need the params key, for example.

Do you have any other examples?

kyrylo: I understand what you want to achieve, but:

1. A Hash must be returned, so we could merge it with the Notice object, so it could be filtered & truncated, if needed
2. We could achieve what you suggest only if we allowed populating the "params" bit of the payload (ignoring context, session and the rest), but how useful would that be? I think we could afford it for the sake of simpler implementation, but we still can't overcome (1)
3. How can we detect if we should append the exception to the Notice object? Currently we check if `to_airbrake` is defined. With your proposed API this won't be possible and we would need to do `notice.params[exception] = exc` for every exception, which doesn't look good. By default, `to_json` on exception object seems to be calling `to_s`, so if the user doesn't define it, then we get a string like:
 `"params":{"exception":"\"MyException\""}` instead of
 `"params":{"exception":{"foo":"bar"}}`. Because of Rails' ActiveSupport your idea would work, but for vanilla Ruby it doesn't work like that, unfortunately.

P.S. This is how the exception would look like

```rb
class MyException < StandardError do
  def initialize(params)
    @params = params
  end
 
  def to_json(*a)
    @params.to_json(*a)
  end
end

ex = MyException.new(foo: 'bar')
```

hale:
This is really great @kyrylo, thanks. Integrating with the branch at the moment - will let you know when I've had a chance to test end to end.

hale:
Looking good - data defined by `to_airbrake` is passed into Airbrake from the custom exception as expected. Appreciate it!

README.md
kyrylo: I am going to merge this today unless you object.

kyrylo:
Awesome, I think this change will go in, but as for the next release, I can't promise it today or tomorrow (although it's possible). So, I'd say sometime next week.

README.md
```diff
@@ -528,6 +529,38 @@ filtered. Modifiable payload includes:
 notice[:params][:my_param] = 'foobar'
 ```
 
+### Custom exception attributes
+
+The library supports custom exception attributes. This is useful if you work
+with custom exceptions, which define non-standard attributes and you can't
+attach any additional data with help of the [`add_filter`](#airbrakeadd_filter)
+API due to the fact that the data isn't available at confuration time yet.
```
urkle: spelling: confuration -> configuration

kyrylo: Thank you, fixed!

PR #115, opened 2016-09-05
config: validate 'environment'

kyrylo:
Our dashboard expects it to be a String. Now we validate it and permit
only allowed types.

I also refactored the config class and extracted validation steps to a
separate class.

### DIFF
--- a/lib/airbrake-ruby/config/validator.rb
+++ b/lib/airbrake-ruby/config/validator.rb
@@ -1,0 +1,73 @@
+module Airbrake
+  class Config
+    ##
+    # Validates values of {Airbrake::Config} options.
+    #
+    # @api private
+    # @since v1.5.0
+    class Validator
+      ##
+      # @return [String]
+      REQUIRED_KEY_MSG = ':project_key is required'.freeze
+
+      ##
+      # @return [String]
+      REQUIRED_ID_MSG = ':project_id is required'.freeze
+
+      ##
+      # @return [String]
+      WRONG_ENV_TYPE_MSG = "the 'environment' option must be configured " \
+                           "with a Symbol (or String), but '%s' was provided: " \
+                           "%s".freeze
+
+      ##
+      # @return [Array<Class>] the list of allowed types to configure the
+      #   environment option
+      VALID_ENV_TYPES = [NilClass, String, Symbol].freeze
+
+      ##
+      # @return [String] error message, if validator was able to find any errors
+      #   in the config
+      attr_reader :error_message
+
+      ##
+      # Validates given config and stores error message, if any errors were
+      # found.
+      #
+      # @param config [Airbrake::Config] the config to validate
+      def initialize(config)
+        @config = config
+        @error_message = nil
+      end
+
+      ##
+      # @return [Boolean]
+      def valid_project_id?
+        valid = @config.project_id.to_i > 0
+        @error_message = REQUIRED_ID_MSG unless valid
+        valid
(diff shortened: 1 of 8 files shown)


kyrylo:
PTAL

lib/airbrake-ruby/config/validator.rb
```diff
@@ -0,0 +1,69 @@
...
+      # found.
+      #
+      # @param config [Airbrake::Config] the config to validate
+      def initialize(config)
+        @config = config
+        @error_message = nil
+      end
+
+      ##
+      # @return [Boolean]
```
shifi: This will not return `boolean`.
It will return `nil` or `Fixnum` unless I'm missing something?

kyrylo: Not sure why you think so. It will return Boolean and the tests prove this:
https://github.com/airbrake/airbrake-ruby/pull/115/files/14231445e58bcae105865ff5512b68c9e51e4c91#diff-c7a89698d88722526dd499d062dca276R6

lib/airbrake-ruby/config/validator.rb
```diff
@@ -0,0 +1,69 @@
...
+      def valid_project_id?
+        valid = @config.project_id.to_i.nonzero?
+        @error_message = REQUIRED_ID_KEY_MSG unless valid
```
shifi: Maybe you could add separate error message for invalid project id, and separate error message for invalid project key?

lib/airbrake-ruby/config/validator.rb
kyrylo: You are correct, `nonzero?` doesn't return boolean. I fixed it by https://github.com/airbrake/airbrake-ruby/pull/115/commits/ba393eb72c43d9e4798c6acd3a701d48ba546189

lib/airbrake-ruby/config/validator.rb
```diff
@@ -0,0 +1,69 @@
...
+        valid = @config.project_key.is_a?(String) && !@config.project_key.empty?
+        @error_message = REQUIRED_ID_KEY_MSG unless valid
+        valid
+      end
+
+      ##
+      # @return [Boolean]
+      def valid_environment?
+        environment = @config.environment
+        valid = VALID_ENV_TYPES.any? { |type| environment.is_a?(type) }
```
shifi: Why `VALID_ENV_TYPES.any?` here?
`valid = VALID_ENV_TYPES.include? @config.environment.class` should work.

kyrylo: 👍

kyrylo: Fixed by https://github.com/airbrake/airbrake-ruby/pull/115/commits/648de0e4d92a9183b903a0d7de6f307f1e1ce5db.

shifi:
Few small comments. Otherwise LGTM 👍

lib/airbrake-ruby/config/validator.rb
kyrylo: Sounds like a good idea, given that now we have the Validator class, which simplifies validations a lot.

Fixed.

lib/airbrake-ruby/config/validator.rb
kyrylo: Had to revert this because it turned out Rails uses a wrapper class for `Rails.env`, which breaks the code you proposed.

PR #118, opened 2016-09-09
keys_filter: filter out individual user keys

### LINKED ISSUE #114: Context keys are not filtered
Configuring with `c.blacklist_keys = [/password/i, 'email', 'name']` doesn't filter user's email and name. `Airbrake::Rack::NoticeBuilder` pulls the values from Warden and adds them to `context`.

I noticed `:context` is not part of `Airbrake::Filters::FILTERABLE_KEYS` https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/filters.rb#L8

I'm guessing I can work around the problem by writing a custom filter, but would be nice if the blacklist applied to `:context`.

kyrylo:
Fixes #114 (Context keys are not filtered)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ Airbrake Ruby Changelog
   ([#120](https://github.com/airbrake/airbrake-ruby/pull/120))
 * Improved support for parsing JRuby backtraces
   ([#119](https://github.com/airbrake/airbrake-ruby/pull/119))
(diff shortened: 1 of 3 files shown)


vmihailenco:
And why they are still not filtered and instead you manually filter only user key?

kyrylo:
We don't want to allow filtering the context payload because it has information, which is important for us (such as notifier info, etc).

PR #125, opened 2016-09-23
notifier: delete deprecated {whitelist,blacklist}_keys methods

### LINKED ISSUE #123: Delete deprecated methods
It's about time to delete some [deprecated methods](https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby.rb#L191-L207).

kyrylo:
Fixes #123 (Delete deprecated methods)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ Airbrake Ruby Changelog
 
 * Added support for blacklisting/whitelisting using procs
   ([#108](https://github.com/airbrake/airbrake-ruby/pull/108))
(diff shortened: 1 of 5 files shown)


shifi:
CI alert 🔴

kyrylo:
Fixed CI. I have refactored old tests because they were relying on that old API (and removed 2 tests specific to those removed methods).

PTAL

kyrylo:
This change will require a MINOR version bump, but I would like to postpone it because we've bumped it recently. I would like to get more fixes/features before merging this PR, so the bump is justified.

shifi:
👍

PR #126, opened 2016-09-29
circle: update rubies

kyrylo:
* Update all TEENY version of Rubies
* Test against JRuby 9.0.5.0
* Test against MRI 2.4.0-preview2
* Stop testing against Rubinius
  This was a hard decision, but Rubinius is least stable Ruby
  implementation. I am not even sure whether anybody uses it in
  production. We often have build failures because of it. The nail in
  the coffin is its current state. Out of thin air our [builds have
  started failing on Rubinius][1]. The reason is that Rubinius wants a
  modern LLVM. I've failed to configure it to use modern LLVM (was
  getting some segfaults). Given little interest in this implementation
  in the Ruby community and the amount of time I've spent to make it
  work, I think moving on would be the best solution.

[1]: https://circleci.com/gh/airbrake/airbrake-ruby/468

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -15,20 +15,22 @@ dependencies:
             ;;
           1)
             rvm-exec 2.0.0-p645 bundle install --path=vendor
-            rvm-exec 2.1.8 bundle install --path=vendor
+            rvm-exec 2.1.10 bundle install --path=vendor
             ;;
           2)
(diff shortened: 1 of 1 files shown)


zefer:
> Stop testing against Rubinius

👍 We can always revisit in the future if there is enough interest.

PR #134, opened 2016-10-19
Send notice time as part of its context

eksoverzero:
Is it not useful to send through the time of the exception? As far as I can tell, otherwise, the exception's time will show in whichever interface as the time it was received by the API.


kyrylo:
`context/time` is not an API compliant key. I recommend you to attach that info to `params`. I've never heard requests for this kind of information, but if you want to log time, then you can attach a custom filter.

```rb
  notice[:params][:time] = Time.now.utc
end
```

PR #143, opened 2016-12-16
Add the promise API

### LINKED ISSUE #138: support return values in async mode
it would be great to get back a url even in async mode, could be done by for example generating a UUID on the client and sending that to the server ... then the client can ask for that UUID later and get the response if any ... this would allow bringing back the user-informer ... and allow linking airbrakes to errors seen in the logs

kyrylo:
Fixes #138 (support return values in async mode)


lib/airbrake-ruby/promise.rb
```diff
@@ -0,0 +1,75 @@
...
+    #
+    # @example
+    #   Airbrake::Promise.new.catch { |error| raise error }
+    #
+    # @yield [error] The error message from the API
+    # @yieldparam error [String]
+    # @return [self]
+    def catch(&block)
+      yield(@rejected_with) if @rejected_with
+      @on_rejected << block
```
grosser: there is no need to store the block if it was already `@rejected_with`

lib/airbrake-ruby/promise.rb
```diff
@@ -0,0 +1,75 @@
...
+    #
+    # @example
+    #   Airbrake::Promise.new.resolve(123)
+    #
+    # @param notice_id [String]
+    # @return [self]
+    def resolve(notice_id)
+      @resolved_with = notice_id
+      @on_resolved.each { |callback| callback.call(notice_id) }
+      self
```
grosser: this is racy and could lead to hooks not getting called ... needs a mutex

```
A: checks that @rejected_with -> false
B: sets @rejected_with
B: calls all callbacks
A: inserts callback
```

grosser:
this looks pretty good and could have a few other usecases :)
... ideally the client would generate the id so nothing has to wait, but I guess that's a much more complicated change ...

grosser:
once this is merged, an optional UserInformerMiddleware can be added to `airbrake` ... or should that go here ?

lib/airbrake-ruby/notifier.rb
```diff
@@ -49,8 +49,9 @@ def initialize(user_config)
     ##
     # @macro see_public_api_method
     def notify(exception, params = {})
-      send_notice(exception, params)
-      nil
+      promise = Airbrake::Promise.new
+      send_notice(exception, params, default_sender, promise)
```
vmihailenco: look like send_notice can create and return Promise

lib/airbrake-ruby/sync_sender.rb
```diff
@@ -20,7 +20,7 @@ def initialize(config)
     # @param [Airbrake::Notice] notice
     # @param [Airbrake::Notice] endpoint
     # @return [Hash{String=>String}] the parsed HTTP response
-    def send(notice, endpoint = @config.endpoint)
+    def send(notice, _promise = nil, endpoint = @config.endpoint)
```
vmihailenco: Looks like you need to resolve promise here too.

lib/airbrake-ruby/promise.rb
```diff
@@ -0,0 +1,75 @@
+module Airbrake
+  ##
+  # Represents a simplified promise object (similar to promises found in
+  # JavaScript), which allows chaining callbacks that are executed when the
+  # promise is either resolved or rejected.
+  #
+  # @see https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise
```
vmihailenco: I think we should model API after https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/promise.rb. E.g. I noticed that Ruby uses resque instead of catch and it makes sense :)

lib/airbrake-ruby/promise.rb
kyrylo: Fixed.

lib/airbrake-ruby/notifier.rb
kyrylo: Done.

lib/airbrake-ruby/sync_sender.rb
kyrylo: Refactored in a new commit.

lib/airbrake-ruby/promise.rb
kyrylo: Renamed and added 2nd link to that resource. It's no big deal, though. Concurrent-ruby aliases to `catch`.

kyrylo:
> once this is merged, an optional UserInformerMiddleware can be added to airbrake ... or should that go here ?

Rack middlewares go to the airbrake gem. As for the middleware itself, I'm not sure we need to ship it with the gem. We could perhaps document this possibility and show a small example.

kyrylo:
PTAL

lib/airbrake-ruby/notifier.rb
```diff
@@ -49,14 +49,14 @@ def initialize(user_config)
...
-      nil
+      send_notice(exception, params, default_sender)
     end
 
     ##
     # @macro see_public_api_method
     def notify_sync(exception, params = {})
-      send_notice(exception, params, @sync_sender)
+      promise = send_notice(exception, params, @sync_sender)
+      promise.value || promise.reason
```
vmihailenco: It looks strange that you return notice id or error, but probably this is common enough in Ruby? Not sure if we can, but I would change it to always be hash with 2 fields:
- id - notice id on success
- error - error message on failure

vmihailenco:
LGTM

README.md
```diff
@@ -562,6 +565,39 @@ Accesses a notice's payload, which can be read or filtered. Payload includes:
...
+promise is resolved whenever the Airbrake API successfully accepts your
+exception.
+
+Yields notice id (the id of an error at Airbrake, discoverable by
+visiting http://airbrake.io/locate/{{notice_id}}) and returns `self`.
+
+```rb
+Airbrake.notify('Oops').then { |notice_id| puts notice_id }
+#=> 000543c7-95ac-06ba-edd6-e81084f4f50b
+```
```
grosser: As an idea: add `#value` that blocks and then returns the value for the UserInformer usecase

lib/airbrake-ruby/promise.rb
```diff
@@ -0,0 +1,99 @@
...
+    ##
+    # Resolves the promise.
+    #
+    # @example
+    #   Airbrake::Promise.new.resolve(123)
+    #
+    # @param notice_id [String]
+    # @return [self]
+    def resolve(notice_id)
+      @mutex.synchronize do
```
grosser: FYI does not need the synchronize since it only reads ... but does not really matter

grosser:
👍

lib/airbrake-ruby/promise.rb
kyrylo: Does it? Won't there be a race condition when thread A sets `@value` and thread B reads it in the `then` method? Or do you say that the synchronisation in the `then` method is enough to cover this case?

README.md
grosser: There is always a stupid solution ... but better get this merged then adding more stuff ...
```Ruby
def value
  value = nil
  Thread.new { promise.then { |e| value = e || false } }
  sleep 0.01 if value.nil?
  value
end
```

lib/airbrake-ruby/promise.rb
grosser: yes, the sync in the then method is enough ... this code just reads the current state, whatever that might be ... so no need to synchronize with anyone

vmihailenco: AFAIK sync is needed on write *and* reads.

kyrylo:
@grosser thanks a bunch for your help here!

PR #145, opened 2017-01-04
config: allow using URL slugs for 'host'

### LINKED ISSUE #144: Can't reach a host with a subpath
Awesome tool!,
Trying to explore the concept of error catching with Airbrake client and with Errbit; Tried succesfuly in local environment with multiple applications and porting on localhost but when i try to deploy on a VPS with separate application with subpaths i can't get the Airbrake-ruby (or Airbrake 5 gem) to reach a subpath using the host attribute in the configuration.

https://github.com/airbrake/airbrake-ruby#host

`config.host = 'http://my-website.com/errbit'`

But if i move my errbit alias to `/` i get it working perfectly. 
Currently when i do `rake airbrake:test` it throws a 404 error. 

So far i've followed the code up to the endpoint method which calls the host attribute, but i am not sure why it won't work while using host + subpath as the entire host

https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/config.rb#L102

Any ideas on how to proceed?
Thank you!

Edit: Uh oh, should i've posted this on airbrake instead of airbrake-ruby

Edit 2: Aha!, Ok, see how interesting. At the end of that endpoint method the class makes a call to `URI.join`, and just figured the following behaviour:

`2.3.1 :003 > URI.join('http://localhost/monitoring/','/api/etc')`
` => #<URI::HTTP http://localhost/api/etc> `

Note that it trashes away the subpath of the first parameter of the .join method, like only accepting the protocol and the domain (and the port). A solution to my problem would be to keep using my errbit under a port :(, or [even better understand how the URI class works](http://apidock.com/ruby/URI/join/class), but as the endpoint method is implemented i see that i have no other option than rewriting this method.

i wonder how can i extend it?

kyrylo:
Fixes #144 (Can't reach a host with a subpath)

### DIFF
--- a/lib/airbrake-ruby/config.rb
+++ b/lib/airbrake-ruby/config.rb
@@ -103,7 +103,7 @@ module Airbrake
       @endpoint ||=
         begin
           self.host = ('https://' << host) if host !~ %r{\Ahttps?://}
-          api = "/api/v3/projects/#{project_id}/notices?key=#{project_key}"
+          api = "api/v3/projects/#{project_id}/notices?key=#{project_key}"
           URI.join(host, api)
         end
     end
@@ -162,11 +162,6 @@ module Airbrake
       __send__("#{option}=", value)
     rescue NoMethodError
       raise Airbrake::Error, "unknown option '#{option}'"
-    end
-
-    def set_endpoint(id, key, host)
-      host = ('https://' << host) if host !~ %r{\Ahttps?://}
-      @endpoint = URI.join(host, "/api/v3/projects/#{id}/notices?key=#{key}")
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/airbrake-ruby/config.rb
```diff
@@ -163,10 +163,5 @@ def set_option(option, value)
     rescue NoMethodError
       raise Airbrake::Error, "unknown option '#{option}'"
     end
-
-    def set_endpoint(id, key, host)
-      host = ('https://' << host) if host !~ %r{\Ahttps?://}
-      @endpoint = URI.join(host, "/api/v3/projects/#{id}/notices?key=#{key}")
-    end
```
kyrylo: Fun fact: this method was unused ever since we released v5.

PR #146, opened 2017-01-06
Remove support for Ruby 1.9.2 & 1.9.3

kyrylo:
* Remove support for 1.9.2 (official support ended on 2014-07-31)
* Remove support for 1.9.3 (official support ended on 2015-02-23)
* Remove support for JRuby 1.7.X (quote from JRuby: "The primary goal of
  1.7 point releases is to fill out any missing compatibility issues
  with Ruby 1.9.3)

### DIFF
--- a/airbrake-ruby.gemspec
+++ b/airbrake-ruby.gemspec
@@ -25,12 +25,11 @@ DESC
   s.files        = ['lib/airbrake-ruby.rb', *Dir.glob('lib/**/*')]
   s.test_files   = Dir.glob('spec/**/*')
 
+  s.required_ruby_version = '>= 2.0'
+
   s.add_development_dependency 'rspec', '~> 3'
   s.add_development_dependency 'rake', '~> 10'
   s.add_development_dependency 'pry', '~> 0'
-
-  # We still support Ruby 1.9.2+, but webmock 2.2.0+ doesn't.
-  s.add_development_dependency 'webmock', '= 2.2.0'
-
+  s.add_development_dependency 'webmock', '~> 2'
   s.add_development_dependency 'benchmark-ips', '~> 2'
 end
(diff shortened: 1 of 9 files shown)


airbrake-ruby.gemspec
```diff
@@ -25,12 +25,12 @@ DESC
   s.files        = ['lib/airbrake-ruby.rb', *Dir.glob('lib/**/*')]
   s.test_files   = Dir.glob('spec/**/*')
 
+  s.required_ruby_version = '>= 2.0'
```
kyrylo: I considered not adding this, but given that we remove a few Ruby 1.9 special cases (especially the one about Exception#cause), this will make the library unusable anyway (I think it will be crashing).

vmihailenco:
Looks like Ruby 1.9 support is just line 10 lines of code... Removing/keeping that code does not look like a big deal, but probably it is worth it. Not sure.

Generally speaking I think this is fine if you release new major version (1.7.0 or probably even 2.0.0).

So LGTM, but probably keep it open for 2-3 days so someone else can show up to convince you not to do it.

kyrylo:
It's harder to support many versions at the same time when testing tools you rely on stop supporting old Rubies. I don't want to pollute our codebase with hacky conditions or lock our dependencies to stone age testing libraries. I had initially wanted to drop 2.0.0 too, but it's still somewhat popular.

The new MINOR release is planned nevertheless (because of Promises or their alternative).

I'll keep it open for a few days, there's no urgency.

vmihailenco:
I think this deserves major version (1.7.0 is fine I guess) so users still can use the notifier by locking their version in Gemspec.

kyrylo:
I agree, but MAJOR according to semver would be `x.0.0`, so `1.7.0` is my planned bump.

vmihailenco:
Yeah, but practically people rarely lock version to `1.*.*`. Usually it is `1.7.*` or `>=1.7.1`.

PR #149, opened 2017-01-17
Don't modify original params

jgaskins:
Ths params hash used by the Notice is modified in a few places. Notably, the component/action keys are deleted, and all hash values get JSON-stringified in the original hash. This PR duplicates the hash to remove that mutation of the original.

I found this because we pass our Rack `env` hash. The modifications to the hash broke one of our Rack middleware entries. We can easily work around that, but I noticed that one of the specs (the one I updated) mentioned "doesn't modify the payload", so I figured this was unintentional behavior.

### DIFF
--- a/lib/airbrake-ruby/notice.rb
+++ b/lib/airbrake-ruby/notice.rb
@@ -55,6 +55,7 @@ module Airbrake
     HOSTNAME = Socket.gethostname.freeze
 
     def initialize(config, exception, params = {})
+      params = params.dup
       @config = config
 
       @payload = {
--- a/spec/notice_spec.rb
+++ b/spec/notice_spec.rb
@@ -27,7 +27,10 @@ RSpec.describe Airbrake::Notice do
           config = Airbrake::Config.new(logger: Logger.new(out))
           notice = nil
 
-          expect { notice = described_class.new(config, ex) }.not_to raise_error
+          # Freeze params so modification will fail
+          params = {}.freeze
+
+          expect { notice = described_class.new(config, ex, params) }.not_to raise_error
           expect(out.string).to match(/#to_airbrake failed:.+Object.+must be a Hash/)
           expect(notice[:params]).to be_empty
         end


kyrylo:
Very good catch! Could you please write a separate test for this behaviour (breaking a hash)? The `component/action` modification was temporary and it should've been deleted by now. I simply forgot about it, but I'll delete them soon.

jgaskins:
> Could you please write a separate test for this behaviour (breaking a hash)?

I froze the hash in the spec so it would fail if the hash was modified. Is that what you mean or would you like a separate spec for it?

For additional clarification, I couldn't figure out where in the `Airbrake.notify` call stack our Rack `env` got changed, but it didn't seem to be in `Notice`. This just seemed to be the best place to `dup` it to keep all of the mutation localized.

kyrylo:
> I froze the hash in the spec so it would fail if the hash was modified. Is that what you mean or would you like a separate spec for it?

The test fails only because `component/action` is mutated. When I remove them it stops failing, so there has to be a better test. Namely, it'd be nice to test what you describe here:

> I found this because we pass our Rack env hash. The modifications to the hash broke one of our Rack middleware entries. 

Just write a separate test that doesn't rely on `component/action`. I'm going to remove them and you could rebase this branch on `master`, so your test will start failing.

kyrylo:
On second thought, in order not to break a semver rule (we try to follow semver), I am going to deprecate this first.

http://semver.org/#how-should-i-handle-deprecating-functionality

I don't think we'll do a major release after deprecating it though, as semver suggests.

kyrylo:
I just merged https://github.com/airbrake/airbrake-ruby/pull/151, which deprecates the behaviour your test relies on.

kyrylo:
I discovered that `params.dup` won't actually duplicate any values, so I tried to write my own PR instead. It turned out to be a non-trivial task because you can't easily `dup` a lot of values. Could you please show me an example of what you mention here:

> I found this because we pass our Rack env hash. The modifications to the hash broke one of our Rack middleware entries.

What kind of modifications are you referring to? If this is your filter, I can recommend fixing it. Unfortunately it'd be tricky to avoid mutations in a good, fast and robust way.

jgaskins:
Right, it doesn't duplicate it all the way down, which is great for memory consumption but probably also kinda misleading. :-) The main thing (for us, at least) is that it stops `params[key] = transform(params[key])`.

>Could you please show me an example of what you mention here:
> > I found this because we pass our Rack env hash. The modifications to the hash broke one of our Rack middleware entries.

> What kind of modifications are you referring to?

One of our middleware layers grabs an auth token from a header and sets `env['user']` to the user object associated with that auth token. This is how our app knows who is making the request.

Another middleware layer fires off logging and metrics tracking for the user object. We noticed that when we made Airbrake calls, `env['user']` became a string containing the JSON representation of that user object rather than the user object itself, causing `NoMethodError`s. Unfortunately, this was happening outside of the scope protected by our error-handling code because it had already caught the error and reported up to Airbrake by that time. :-)

When we replaced `Airbrake.notify(e, env)` with `Airbrake.notify(e, env.dup)`, this error stopped happening. We don't have filters or any integration with Airbrake in this app except for `Airbrake.notify` calls.

kyrylo:
Thanks for the description!

Do you have a backtrace of that error? I'm not sure why `airbrake-ruby` would be modifying params to be honest. Yes, we had that code that `delete`'s keys, however as of v1.7.0 it was removed (therefore your tests written here should be failing now). We call `.to_json` on your user object, but we don't actually overwrite `params` anywhere, unless I missed something 😕

PR #155, opened 2017-01-20
notice: prevent mutation of params

kyrylo:
Replaces #149 (Don't modify original params)

### DIFF
--- a/lib/airbrake-ruby/notice.rb
+++ b/lib/airbrake-ruby/notice.rb
@@ -63,6 +63,8 @@ module Airbrake
     attr_reader :stash
 
     def initialize(config, exception, params = {})
+      params = deep_dup(params)
+
       @config = config
 
       @payload = {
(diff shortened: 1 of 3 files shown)


lib/airbrake-ruby/notice.rb
```diff
@@ -240,6 +242,15 @@ def extract_custom_attributes(exception)
         )
       end
     end
+
+    def deep_dup(params, level = 1)
+      level = 0 if level > 3
```
vmihailenco: why this line? can't you instead `return params.dup if level > 3`

kyrylo: That'll work.

I spotted more serious issues with the approach (test failures).

vmihailenco:
This always generates some garbage to dup params even if filters don't modify params. Instead you could try to carefully write filters so params are copied only if modification is needed and require users to do the same.

kyrylo:
Closing because we discovered this is probably not the best approach to resolve the issue.

PR #163, opened 2017-02-06
Errbit fix

rafalpetryka:
Errbit by default has project_id equals -1

Airbrake example config
```
  config.host = 'some_url'
  config.project_id = -1
  config.project_key = 'some_project_key'
end
```

### DIFF
--- a/lib/airbrake-ruby/config/validator.rb
+++ b/lib/airbrake-ruby/config/validator.rb
@@ -43,7 +43,7 @@ module Airbrake
       ##
       # @return [Boolean]
       def valid_project_id?
-        valid = @config.project_id.to_i > 0
+        valid = @config.project_id.to_i != 0
         @error_message = REQUIRED_ID_MSG unless valid
         valid
       end


kyrylo:
Hi, thanks for the PR! Does it mean that when the notifier is configured with `-1`, and when I execute `Airbrake.notify`, then Errbit will display that error somewhere?

rafalpetryka:
I don't know about Airbrake.notify, but airbrake rake test task show error in errbit if I set -1 as project_id

⁣Rafał Petryka
533488388​

kyrylo:
Sorry, I'm not following. Why is this a problem?

rafalpetryka:
Errbit should have project_id equals -1 in airbrake config. In your code you have an validation on project_id which i corrected in pull request

kyrylo:
Cool. So does it mean Errbit doesn't support `project_id` at all?
If that's the case, then does mean nobody can use Errbit ever since that check was added?

P. S. I thought it was Airbrake API-compliant :)

kyrylo:
Friendly ping.

rafalpetryka:
Ok, I googled and found some information..
Errbit fixed this error
https://github.com/errbit/errbit/issues/1102
Sorry I still have errbit 0.6.0, this problem is solved in 0.6.1 :/

kyrylo:
Thanks for the details!
I have to close this because the fix is already released. I hope it's not a big deal to upgrade from Errbit v0.6.0 to v0.6.1.

The code in this PR will accept values such as `"-123"`, which isn't valid. Besides, as I mentioned earlier, `-1` was never compatible with our Notice API v3.

If you have any other issues when using the gem with Errbit, let me know and we'll figure something out, too 👍

rafalpetryka:
Thx 😃

PR #168, opened 2017-02-20
airbrake-ruby: deprecate `notifier_name` argument

### LINKED ISSUE #165: Get rid of third argument in public API methods
The third argument is `notifier_name`. I never really liked the way it's designed, but it's was simple in terms of implementation. Reminder, this is the old API:

```rb
Airbrake.notify('oops', params, :project_a)
# ...
Airbrake.any_function_call(any_args, :project_a)
```

Instead, I'd like to decouple `:project_a` from method signatures. Proposed way:

```rb
Airbrake[:project_a].notify('oops', params)
```

The old way of calling Airbrake must also be supported: `Airbrake.notify`.

kyrylo:
Partially resolves #165 (Get rid of third argument in public API
methods)

### DIFF
--- a/lib/airbrake-ruby.rb
+++ b/lib/airbrake-ruby.rb
@@ -53,7 +53,7 @@ require 'airbrake-ruby/notifier'
 #
 #   # Send an exception via other configured notifier.
 #   params = {}
-#   Airbrake.notify('Oops', params, :my_other_project)
+#   Airbrake[:my_other_project].notify('Oops', params)
 #
 # @see Airbrake::Notifier
 # @since v1.0.0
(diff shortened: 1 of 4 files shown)


lib/airbrake-ruby.rb
```diff
@@ -139,7 +151,16 @@ def configure(notifier = :default)
...
+    def notify(exception, params = {}, notifier = (no_arg = true && :default))
+      unless no_arg
+        warn(
+          "#{LOG_LABEL} `Airbrake.notify` method signature was changed. " \
+          "Passing `notifier_name` is deprecated and will be removed in the " \
+          "next MAJOR release.\n" \
+          "Use `Airbrake[:#{notifier}]` to access the :#{notifier} notifier " \
+          "and call same methods directly on it."
+        )
+      end
```
shifi: Move this to method and re-use it?

kyrylo: 👍

kyrylo:
PTAL

grosser:
would it be simpler to make it part of the options ?
`Airbrake.notify(foo, notifier: :bar)`
(never used this feature)

kyrylo:
@grosser IMO that would be an overkill because we don't have any other arguments to pass.

grosser:
I'm fine with either ... don't really care ... only thing that I could use is if `Airbrake.configuration` started working again.

kyrylo:
@grosser please file a new issue about this.

PR #170, opened 2017-02-20
enforce code coverage and show what is not covered

grosser:
[single_cov](https://github.com/grosser/single_cov) makes test run fail on uncovered code and shows exactly where coverage is missing.

```
...

Finished in 2.07 seconds (files took 0.4644 seconds to load)
14 examples, 0 failures

Randomized with seed 49008

lib/airbrake-ruby.rb new uncovered lines introduced (2 current vs 0 configured)
Lines missing coverage:
lib/airbrake-ruby.rb:223
lib/airbrake-ruby.rb:239
```

### DIFF
--- a/airbrake-ruby.gemspec
+++ b/airbrake-ruby.gemspec
@@ -33,4 +33,5 @@ DESC
   s.add_development_dependency 'webmock', '~> 2.3'
   s.add_development_dependency 'benchmark-ips', '~> 2'
   s.add_development_dependency 'rubocop', '~> 0.47'
+  s.add_development_dependency 'single_cov', '~> 0'
 end
--- a/spec/airbrake_spec.rb
+++ b/spec/airbrake_spec.rb
@@ -1,5 +1,7 @@
 require 'spec_helper'
 
+SingleCov.covered! file: 'lib/airbrake-ruby.rb', uncovered: 2
+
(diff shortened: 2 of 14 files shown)


kyrylo:
Thank you for your PR. 

I'm afraid I'm inclined not to merge this because we don't strive for 100% code coverage (some things are too trivial to be tested) and with this tool we are forced to do that, otherwise the build fails. I would also like to keep the list of deps (even dev deps) as small as possible because the more we have, the slower Circle runs. Also, as of now there's no real need to have a tool like single_cov (although under some circumstances it is useful indeed).

Last but not least, I can see you use Ruby's [`coverage`](https://ruby-doc.org/stdlib-2.4.0/libdoc/coverage/rdoc/Coverage.html) from stdlib. It is considered to be experimental:

> Coverage provides coverage measurement feature for Ruby. This feature is experimental, so these APIs may be changed in future.

grosser:
The install overhead is minimal ... 1-5 seconds I'd guess.
It does not force anyone to add coverage, but it shows where coverage is
missing, which may be unexpected.
It found 7 tests that had uncovered code, so I'd say it is helpful / would
be nice to add tests for these areas.
Coverage api has not changed since 1.9 and even if it were to change it
would be easy to either update or remove that gem.

PR #173, opened 2017-02-23
Add support for severity

kyrylo:


vmihailenco:
I think that technically this is fine, but practically almost none will manually set notice severity. And even if they are willing to do it then doing it via UI should be much more easier.

What about adding integration with https://ruby-doc.org/stdlib-2.1.0/libdoc/logger/rdoc/Logger.html and using Logger severity?

kyrylo:
Take a look at https://github.com/airbrake/airbrake/pull/674

kyrylo:
If any of the users are interested in this kind of API, then speak up.
This PR simply introduces a shortcut for setting severities.

You can do this any time to set your own severity:

```rb
notice = Airbrake.build_notice('oops')
notice[:context][:severity] = 'critical'
Airbrake.notify(notice)
```

PR #180, opened 2017-03-17
resolve symlinked roots since that is what is shown in backtraces

grosser:
Many of our apps are deployed in symlinked root directories, for example: `/deploys/foo/current` which points to `/deploys/foo/releases/20170316121212`. Rubies backtraces always include the fully resolved path, for example '/deploys/foo/releases/20170316121212/app/models/foo.rb:12' ... so removing the symlinked root from it does not work and results in us reporting errors that cannot be grouped together.

Fix: resolve root to it's real path and the matching works.

(Tested that realpath works with strings and Pathnames)

### DIFF
--- a/lib/airbrake-ruby/config.rb
+++ b/lib/airbrake-ruby/config.rb
@@ -96,9 +96,9 @@ module Airbrake
       self.blacklist_keys = []
       self.whitelist_keys = []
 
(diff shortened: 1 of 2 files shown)


kyrylo:
That makes sense. 👍

PR #192, opened 2017-04-15
Fix capistrano airbrake:deploy task

akicho8:
```ruby
URI.join("http://example.net/myapp/", "/api").to_s # => "http://example.net/api"
URI.join("http://example.net/myapp/", "api").to_s  # => "http://example.net/myapp/api"

```
`/myapp/` disappears when there is a slash

### DIFF
--- a/spec/notifier_spec.rb
+++ b/spec/notifier_spec.rb
@@ -539,25 +539,25 @@ RSpec.describe Airbrake::Notifier do
       "https://airbrake.io/api/v4/projects/#{project_id}/deploys?key=#{project_key}"
     end
 
-    it "sends a request to the deploy API" do
+    before do
       stub_request(:post, deploy_endpoint).to_return(status: 201, body: '{"id":"123"}')
+    end
+
+    it "sends a request to the deploy API" do
       @airbrake.create_deploy({})
       expect(a_request(:post, deploy_endpoint)).to have_been_made.once
     end
 
-    describe "when a host contains subdirectories" do
-      let(:deploy_host) do
-        "https://airbrake.io/subdir/"
-      end
+    context "when a host contains paths" do
+      let(:deploy_host) { "https://example.net/errbit/" }
 
       let(:deploy_endpoint) do
         "#{deploy_host}api/v4/projects/#{project_id}/deploys?key=#{project_key}"
       end
 
       it "sends a request to the deploy API" do
-        @airbrake.instance_variable_get(:@config).host = deploy_host
-        stub_request(:post, deploy_endpoint).to_return(status: 201, body: '{"id":"123"}')
-        @airbrake.create_deploy({})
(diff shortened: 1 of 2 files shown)


repus:
Open

kyrylo:
Hey @akicho8! 
This looks good. Nice observation with `@config.host`.
I'm happy to merge this, but we need to add a test.

spec/notifier_spec.rb
```diff
@@ -544,5 +544,22 @@ def to_json(*)
       @airbrake.create_deploy({})
       expect(a_request(:post, deploy_endpoint)).to have_been_made.once
     end
+
+    describe "when a host contains subdirectories" do
```
kyrylo: This should be `context`, not `describe`.

spec/notifier_spec.rb
```diff
@@ -544,5 +544,22 @@ def to_json(*)
...
+      let(:deploy_host) do
+        "https://airbrake.io/subdir/"
```
kyrylo: Airbrake doesn't use paths for hosts. I assume the fix is needed for Errbit.

1. Would you mind changing `airbrake.io` to an Errbit URL, so it's more clear
2. `subdir` is a wrong term here, I would call this `subpath` since it has nothing to do with directories

spec/notifier_spec.rb
```diff
@@ -544,5 +544,22 @@ def to_json(*)
...
+      end
+
+      let(:deploy_endpoint) do
+        "#{deploy_host}api/v4/projects/#{project_id}/deploys?key=#{project_key}"
+      end
+
+      it "sends a request to the deploy API" do
+        @airbrake.instance_variable_get(:@config).host = deploy_host
+        stub_request(:post, deploy_endpoint).to_return(status: 201, body: '{"id":"123"}')
```
kyrylo: This duplicates line 543, it can be moved to a before block:

```ruby
describe "#create_deploy" do
  before do
    stub_request(:post, deploy_endpoint).to_return(status: 201, body: '{"id":"123"}')
  end
...
```

spec/notifier_spec.rb
```diff
@@ -544,5 +544,22 @@ def to_json(*)
...
+      let(:deploy_host) do
+        "https://airbrake.io/subdir/"
+      end
+
+      let(:deploy_endpoint) do
+        "#{deploy_host}api/v4/projects/#{project_id}/deploys?key=#{project_key}"
+      end
+
+      it "sends a request to the deploy API" do
+        @airbrake.instance_variable_get(:@config).host = deploy_host
```
kyrylo: `instance_variable_get` should be avoided when possible. In our case we can create a new notifier and configure it accordingly:

```ruby
described_class.new(airbrake_params.merge(host: deploy_host))
```

kyrylo:
👍

akicho8:
Thanks.
But errbit did not correspond to v4 api. :sweat_smile:

PR #194, opened 2017-04-24
Implemented support for thread context

### LINKED ISSUE #187: Save context of a current thread
Lets assume we have a complex logic, which has multiple steps, and you can't access previous variables that affects on the current state of logic where an exception happened. Right now it's not possible to find out how it came into current state, except to go through the logger for the same time. Would be nice to save context in a `Thread.current` and use it inside the `Notifier`.

kyrylo:
Fixes #187 (Save context of a current thread)

### DIFF
--- a/lib/airbrake-ruby.rb
+++ b/lib/airbrake-ruby.rb
@@ -22,6 +22,7 @@ require 'airbrake-ruby/filters/keys_whitelist'
 require 'airbrake-ruby/filters/keys_blacklist'
 require 'airbrake-ruby/filter_chain'
 require 'airbrake-ruby/notifier'
+require 'airbrake-ruby/thread_context'
 
 ##
 # This module defines the Airbrake API. The user is meant to interact with
@@ -239,6 +240,24 @@ module Airbrake
     # @return [void]
     def create_deploy(deploy_params)
       @notifiers[:default].create_deploy(deploy_params)
+    end
+
+    ##
+    # Returns current thread context. Can be used to set new thread context
+    # values. The thread context is cleared on every `notify/notify_sync` call,
+    # and also and when the current thread `join`'s. The values assigned to the
+    # thread context are passed to the `params` field of a notice.
+    #
+    # @example
+    #   Airbrake.thread_context[:bingo] = :bango
+    #
+    #   # This call will deliver `{ :bingo => :bango }` to the dashboard.
+    #   Airbrake.notify('Oops')
+    #
+    # @since v2.1.0
+    # @return [Airbrake::ThreadContext] the thread context of
+    def thread_context
+      @notifiers[:default].thread_context
     end
   end
 end
--- a/lib/airbrake-ruby/filter_chain.rb
+++ b/lib/airbrake-ruby/filter_chain.rb
@@ -45,7 +45,7 @@ module Airbrake
 
     ##
     # @param [Airbrake::Config] config
-    def initialize(config)
+    def initialize(config, thread_context)
       @filters = []
       @keys_filters = []
 
@@ -55,6 +55,7 @@ module Airbrake
 
       root_directory = config.root_directory
       add_filter(root_directory_filter(root_directory)) if root_directory
+      add_filter(thread_context_filter(thread_context))
     end
 
     ##
@@ -87,6 +88,15 @@ module Airbrake
             next unless (file = frame[:file])
             file.sub!(/\A#{root_directory}/, '[PROJECT_ROOT]'.freeze)
           end
+        end
+      end
+    end
+
+    def thread_context_filter(thread_context)
+      proc do |notice|
+        unless (ctx = thread_context.to_h).empty?
+          notice[:params][:thread_context] = ctx
+          thread_context.clear
         end
       end
     end
--- a/lib/airbrake-ruby/thread_context.rb
+++ b/lib/airbrake-ruby/thread_context.rb
@@ -1,0 +1,56 @@
+module Airbrake
+  ##
+  # A class that is responsible for storing Airbrake Ruby thread context. A
+  # thread context is a hash of arbitrary objects to be sent to Airbrake. The
+  # implementation of the class relies on thread local variables.
+  #
+  # @since 2.1.0
+  class ThreadContext
+    ##
+    # @return [String] the prefix of an Airbrake Ruby thread local variable
+    CONTEXT_PREFIX = 'airbrake_ruby_context'.freeze
+
+    def initialize(name)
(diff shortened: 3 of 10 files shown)


lib/airbrake-ruby.rb
```diff
@@ -240,5 +241,23 @@ def close
...
+    #
+    # @example
+    #   Airbrake.thread_context[:bingo] = :bango
+    #
+    #   # This call will deliver `{ :bingo => :bango }` to the dashboard.
+    #   Airbrake.notify('Oops')
+    #
+    # @since v2.1.0
+    # @return [Airbrake::ThreadContext] the thread context of
+    def thread_context
```
phstc: @kyrylo `.thread_` feels like "leaking" implementation details to the client. WDYT about only `.context`?

lib/airbrake-ruby/filter_chain.rb
```diff
@@ -90,5 +91,14 @@ def root_directory_filter(root_directory)
         end
       end
     end
+
+    def thread_context_filter(thread_context)
+      proc do |notice|
+        unless (ctx = thread_context.to_h).empty?
+          notice[:params][:thread_context] = ctx
+          thread_context.clear
```
phstc: @kyrylo so this guarantees that after a `notify` whatever that was in the `thread_context` gets cleared out?

lib/airbrake-ruby/thread_context.rb
```diff
@@ -0,0 +1,56 @@
...
+  #
+  # @since 2.1.0
+  class ThreadContext
+    ##
+    # @return [String] the prefix of an Airbrake Ruby thread local variable
+    CONTEXT_PREFIX = 'airbrake_ruby_context'.freeze
+
+    def initialize(name)
+      @context_key = "#{CONTEXT_PREFIX}_#{name}"
+      clear
```
phstc: @kyrylo does it need to be cleared out during the initialization?

lib/airbrake-ruby.rb
kyrylo: I am worried that `.context` will clash with our naming for `context` payload (`notice[:context]`), hence that's why I chose `thread_context`. I can see how it's leaking, though. Any other ideas? Do you happen to know if anyone else besides Honeybadger have this feature (and how it's called there)?

lib/airbrake-ruby/filter_chain.rb
kyrylo: Not exactly. Yes, this will clear after each `notify`, however internally it'll actually happen before  notifying.

lib/airbrake-ruby/thread_context.rb
kyrylo: Every notifier will have its own space. This is get constructed when a Notifier is initialised. We call `clear` here to set the initial hash. The same notifier can't be initialised twice, so this should be safe, unless I missed something.

vmihailenco:
And why we don't just "Save context of a current thread"? E.g. `notice.params.thread_vars = Thread.current.all_vars`

kyrylo:
> And why we don't just "Save context of a current thread"? E.g. notice.params.thread_vars = Thread.current.all_vars

I find this solution acceptable and I didn't guess to suggest it.

```ruby
  thread_variables = {}

  Thread.current.thread_variables.each do |key|
    thread_variables[key] = Thread.current.thread_variable_get(key)
  end

  notice[:params][:thread_variables] = thread_variables 
end
```

@phstc this is obviously a simpler solution (and you can already use it with any airbrake-ruby version). Will this work for you?

phstc:
@kyrylo LGTM I will try it out this week (most likely weekend).

lib/airbrake-ruby.rb
phstc: @kyrylo hm `.context` looks perfectly, but I see what you are saying. Maybe `global_context`, I dunno TBH. I don't know other similar tools using it. But I guarantee this is super super useful.

kyrylo:
Closed in favour of https://github.com/airbrake/airbrake-ruby/pull/198.

phstc:
hi @kyrylo 

I was going to test this "for-real", but I got lost. Do you have docs for #198?

kyrylo:
@phstc save anything you want to current thread as a thread or a fiber local and it'll be reported by Airbrake automatically:

```ruby
Thread.current.thread_variable_set(:thing, value)
Thread.current[:thing] = value
```

Basically, all this API that I wrote in this PR already exists, so in #198 airbrake-ruby leverages it without creating an unnecessary abstraction.

Does it make sense now?

phstc:
@kyrylo hm I have mixed feelings with it. But I guess it will work.

Mixed feelings:
- `notify` will re-send thread `params` if called twice or more within the same thread
- `notify` may send `params` it shouldn't for users using already using thread params for something else
- [This thread problem](https://github.com/steveklabnik/request_store#the-problem)

kyrylo:
> notify will re-send thread params if called twice or more within the same thread

I'm happy that we're able to provide some happy medium. We don't overcomplicate our library API, at the same time we provide support for this feature. Obviously, you will need to do the cleanup yourself. So far, I'm inclined to leave this as is

> notify may send params it shouldn't for users using already using thread params for something else

I'd be happy to revisit this with a concrete example (expected A, saw B). This way I could think of how to improve the current "API" that we have.

> This thread problem

Unfortunately, I don't see what this problem is about. It seems to be a superficial description.

---

Nobody has had a chance to use this yet because I haven't updated the airbrake gem. I want to release it and see how it's perceived before making another iteration.

phstc:
> Unfortunately, I don't see what this problem is about. It seems to be a superficial description.

@kyrylo this is a problem with threaded servers, if you `Thread.current[:something_specific_for_a_request_A] = value`, most likely a `request_B` will also have that key & value, since a threaded server will re-use the Thread.

For those using airbrake with web apps, usually the `params` would be erased/set on a request basis. It's how it works with Honeybadger and its "persisted" `context`.

---

Anyway, I think this will help for what I need. Not ideally like `Honeybadger.context`, but it's something. Thanks for looking into it 🍻

dmitry:
Is there are a way to improve PR and merge it with master?

kyrylo:
@dmitry the approach that was aimed to replace this PR turned out to be unsuccessful, so yes, I'm going to proceed with this PR.

kyrylo:
After some considerations, we're going to proceed with https://github.com/airbrake/airbrake-ruby/pull/224 instead. You can put your "context" variables to the current thread and it'll be sent to Airbrake automatically.

PR #198, opened 2017-04-26
Send thread information by default

### LINKED ISSUE #187: Save context of a current thread

kyrylo:
Fixes #187 (Save context of a current thread)
Replaces #194

Instead of providing an abstraction over `Thread.thread_variable_set`,
simply send everything we know about current thread.

### DIFF
--- a/lib/airbrake-ruby/filters/thread_filter.rb
+++ b/lib/airbrake-ruby/filters/thread_filter.rb
@@ -1,0 +1,50 @@
+module Airbrake
+  module Filters
+    ##
+    # Attaches thread & fiber local variables along with general thread
+    # information.
+    class ThreadFilter
+      def call(notice)
+        th = Thread.current
+        thread_info = {}
+
+        if (vars = thread_variables(th)).any?
+          thread_info[:thread_variables] = vars
+        end
+
+        if (vars = fiber_variables(th)).any?
+          thread_info[:fiber_variables] = vars
+        end
+
+        # Present in Ruby 2.3+.
+        if th.respond_to?(:name) && (name = th.name)
+          thread_info[:name] = name
+        end
+
+        add_thread_info(th, thread_info)
+
+        notice[:params][:thread] = thread_info
+      end
+
+      private
+
+      def thread_variables(th)
+        th.thread_variables.map.with_object({}) do |var, h|
+          h[var] = th.thread_variable_get(var).inspect
+        end
+      end
+
+      def fiber_variables(th)
+        th.keys.map.with_object({}) { |key, h| h[key] = th[key].inspect }
+      end
+
+      def add_thread_info(th, thread_info)
+        thread_info[:self] = th.inspect
+        thread_info[:group] = th.group.list.map(&:inspect)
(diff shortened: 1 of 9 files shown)


lib/airbrake-ruby/filters/thread_filter.rb
```diff
@@ -0,0 +1,56 @@
...
+        if (vars = thread_variables(th)).any?
+          thread_info[:thread_variables] = vars
+        end
+
+        if (vars = fiber_variables(th)).any?
+          thread_info[:fiber_variables] = vars
+        end
+
+        if (alive = th.alive?)
+          thread_info[:alive?] = alive
```
vmihailenco: How current thread can possibly be not alive?

lib/airbrake-ruby/filters/thread_filter.rb
```diff
@@ -0,0 +1,56 @@
...
+      def fiber_variables(th)
+        th.keys.map.with_object({}) { |key, h| h[key] = th[key].inspect }
+      end
+
+      def add_thread_info(th, thread_info)
+        thread_info[:self] = th.inspect
+        thread_info[:group] = th.group.list.map(&:inspect)
+        thread_info[:pending_interrupt?] = th.pending_interrupt?
+        thread_info[:priority] = th.priority
+        thread_info[:stop?] = th.stop?
```
vmihailenco: why stop is here and alive is not? how current thread can be dead?

lib/airbrake-ruby/filters/thread_filter.rb
kyrylo: It can't. Removed.

lib/airbrake-ruby/filters/thread_filter.rb
kyrylo: It can't. Removed.

kyrylo:
PTAL

spec/notifier_spec.rb
```diff
@@ -154,9 +154,16 @@ def expect_a_request_with_headers(headers)
...
+            /"params":{"bingo":\["bango"\],"bongo":"bish".*}/
           )
         end
+
+        # it "features 'params/thread'" do
+        #   expect_a_request_with_body(
+        #     /"params":{.*"thread":.*}/
+        #   )
+
+        # end
```
vmihailenco: overlook?

kyrylo: Yeah, forgot to uncomment. Thanks!

kshnurov:
@vmihailenco @kyrylo how do I disable this filter? Adding one more to remove thread variables is a stupid approach.

kyrylo:
There's no other way at the moment. I created https://github.com/airbrake/airbrake-ruby/issues/290 to talk about this.

PR #205, opened 2017-05-02
filter_chain: refactor code

kyrylo:
* filter_chain: rename 'keys_filters' to 'post_filters' 
* filter_chain: sort filters by weight

### DIFF
--- a/lib/airbrake-ruby/filter_chain.rb
+++ b/lib/airbrake-ruby/filter_chain.rb
@@ -8,11 +8,6 @@ module Airbrake
   # @since v1.0.0
   class FilterChain
     ##
-    # @return [String] the namespace for filters, which are executed first,
-    #   before any other filters
-    LIB_NAMESPACE = '#<Airbrake::'.freeze
-
-    ##
     # @return [Array<Class>] filters to be executed first
     DEFAULT_FILTERS = [
       Airbrake::Filters::SystemExitFilter,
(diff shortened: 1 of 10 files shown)


vmihailenco:
I never saw prelude+postlude names in APIs. Usually before/after & pre&post names are used.

Overall you have prelude & postlude but nothing in between which does not make much sense.

lib/airbrake-ruby/filter_chain.rb
```diff
@@ -45,11 +45,11 @@ def initialize(config)
...
     def add_filter(filter)
-      return @keys_filters << filter if KEYS_FILTERS.include?(filter.class)
-      return @filters << filter unless filter.to_s.start_with?(LIB_NAMESPACE)
+      return @postlude_filters << filter if POSTLUDE_FILTERS.include?(filter.class)
+      return @prelude_filters << filter unless filter.to_s.start_with?(LIB_NAMESPACE)
 
-      i = @filters.rindex { |f| f.to_s.start_with?(LIB_NAMESPACE) }
-      @filters.insert(i + 1, filter) if i
+      i = @prelude_filters.rindex { |f| f.to_s.start_with?(LIB_NAMESPACE) }
+      @prelude_filters.insert(i + 1, filter) if i
```
vmihailenco: This function looks unnecessary cryptic.

kyrylo:
Let's go with `before/after`. How about leaving `@filters` but renaming `@keys_filters` to `@post_filters`?

lib/airbrake-ruby/filter_chain.rb
```diff
@@ -59,7 +59,7 @@ def add_filter(filter)
     # @param [Airbrake::Notice] notice The notice to be filtered
     # @return [void]
     def refine(notice)
-      (@filters + @keys_filters).each do |filter|
+      (@prelude_filters + @postlude_filters).each do |filter|
```
vmihailenco: I guess this creates new array each time notice is filtered?

vmihailenco:
>How about leaving @filters but renaming @keys_filters to @post_filters?

Sounds better.

lib/airbrake-ruby/filter_chain.rb
kyrylo: I left only one array. We now `insert` and calculate indices.

lib/airbrake-ruby/filter_chain.rb
kyrylo: Refactored.

kyrylo:
PTAL

lib/airbrake-ruby/filter_chain.rb
```diff
@@ -45,11 +54,12 @@ def initialize(config)
...
-      return @filters << filter unless filter.to_s.start_with?(LIB_NAMESPACE)
-
-      i = @filters.rindex { |f| f.to_s.start_with?(LIB_NAMESPACE) }
-      @filters.insert(i + 1, filter) if i
+      if filter.to_s.start_with?(LIB_NAMESPACE)
+        @filters.insert(@offset_left, filter)
+        @offset_left += 1
+      else
+        @filters.insert(@filters.size - @offset_right, filter)
+      end
```
vmihailenco: This is still ugly and unreadable. What is the problem with adding optional `weight` property to filters (with default value 0) and sort by `filter.weight`? It should not be more than 10 lines of code.

kyrylo: Sure, let's try. I don't think there's a problem. Filter weights will give a lot of flexibility, but nobody has requested this flexibility so far.

kyrylo:
All library filters have a weight of 0. The blacklist/whitelist filters have a weight of 2. User defined filters will have a weight of 1.

PTAL:

kyrylo:
I'm also not documenting this because nobody has requested the feature. So far it's purely for internal use.

vmihailenco:
Generally LGTM

There are 2 moderately confusing things:
- filters with bigger weight are run later but I expect the opposite
- the default is 1, but I expect 0. It is especially confusing to see that you set optional field `weight = 0`.

kyrylo:
If the default is 0, then keys filters will need to have a weight of `-1`, user filters will be `0` (default), library filters will be `1`. Sounds good?

vmihailenco:
Yes, except that I will go with -100, 0, 100 to give some room to insert filters in between. You can go even further and use `-100, -90, 0 (default), 100, 110, 120`, i.e. don't use same weight twice.

PR #216, opened 2017-05-12
commonize method NilNotifier.

vikke:
commonize method interface NilNotifier and Notifier.


lib/airbrake-ruby/notifier.rb
```diff
@@ -155,11 +155,11 @@ def clean_backtrace
   class NilNotifier
     def notify(_exception, _params = {}); end
```
kyrylo: For consistency, it would be nice to add `&block` here, too.

vikke: > For consistency, it would be nice to add &block here, too.

done it.

PR #222, opened 2017-05-15
notice: rescue SystemStackError and JSONify inside ThreadFilter

kyrylo:
Hopefully fixes airbrake/airbrake#743
(Airbrake.notify() broken)

Most likely this error comes from `ThreadFilter`.  Sometimes, Rails'
`active_support` gets confused how to parse certain objects.  More info
here: #208

Unfortunately, we can't rely on the `rescue` inside the Notice class
because if a thread variable fails to be converted to JSON, the whole
notice will be rejected. Thread variables live long enough, so if there
a "bad" variable sticks to a thread, then all further notices would be
rejected as well.

To solve the problem and send thread variables nevertheless,
we perform an [ugly] conversion to JSON inside `ThreadFilter`, with
individual `rescue`'s.

### DIFF
--- a/lib/airbrake-ruby/filters/thread_filter.rb
+++ b/lib/airbrake-ruby/filters/thread_filter.rb
@@ -8,16 +8,6 @@ module Airbrake
       # @return [Integer]
       attr_reader :weight
 
-      ##
-      # @return [Array<Symbol>] the list of ignored fiber variables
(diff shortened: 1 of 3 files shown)


chuckphillips:
It resolved the issue for us. Thank you.

kyrylo:
Replaced by https://github.com/airbrake/airbrake-ruby/pull/223.

PR #223, opened 2017-05-17
Delete ThreadFilter

kyrylo:
Fixes https://github.com/airbrake/airbrake/issues/743
(Airbrake.notify() broken)
Replaces: https://github.com/airbrake/airbrake-ruby/pull/222

ThreadFilter was introduced in v2.1.0. Unfortunately, it tunred out to
be a mistake. Prevalent number of users use airbrake-ruby with Rails,
which includes `active_support/json`. It monkey-patches all objects with
its own methods for JSON conversion. `active_support/json` is incapable
of converting recursive data structures and raises SystemStackError.

The issue is extensively described by (and linked issues):
https://github.com/airbrake/airbrake/issues/743

It's too dangerous to have ThreadFilter because of its unpredicted
results, so I am removing it. This was an answer to the thread context
feature request: https://github.com/airbrake/airbrake-ruby/pull/194

I think we need to reopen and merge that instead.


vmihailenco:
And why https://github.com/airbrake/airbrake-ruby/pull/222 is closed?

kyrylo:
The code there doesn't fix anything. Our user tested it and I agree with the rationale: https://github.com/airbrake/airbrake/issues/743#issuecomment-301916950

I don't see any better solution than this PR so far. I like how `ThreadFilter` it works in theory, but in reality it doesn't work as expected.

vmihailenco:
You could collect only primitive types in thread context, i.e. string numbers, array & hashes, but not classes.

kyrylo:
Replaced by https://github.com/airbrake/airbrake-ruby/issues/224

PR #224, opened 2017-05-17
filters/thread_filter: whitelist allowed variable classes

kyrylo:
Fixes https://github.com/airbrake/airbrake/issues/743
(Airbrake.notify() broken)

Replaces: https://github.com/airbrake/airbrake-ruby/pull/223

This change makes `ThreadFilter` less useful, but more robust.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@ Airbrake Ruby Changelog
 
 ### master
 
+* Fixed bug in `ThreadFilter`, when it attaches an object, which can't be dumped
+  to JSON. As result, `ThreadFilter` has become stricter: it only allows
+  instances of whitelisted classes (primitives)
+  ([#224](https://github.com/airbrake/airbrake-ruby/pull/224))
+
 ### [v2.2.3][v2.2.3] (May 11, 2017)
 
 * Fixed bug in keys filters while trying to filter a non Symbol/String key when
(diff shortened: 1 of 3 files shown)


phstc:
> This change makes ThreadFilter less useful, but more robust.

🐼 

_ misses `Honeybadger.context`_

kyrylo:
You could do the following:

```ruby
Thread.current[:foo] = 42
Airbrake.notify('oops').then { Thread.current[:foo] = nil } 
```

If you want to use `request_store`, it also should work, because `request_store` relies on `Thread.current`, too. Therefore:

```ruby
RequestStore.store[:foo] = 42
Airbrake.notify('oops').then { RequestStore.store[:foo] = nil } 
```

kyrylo:
Merging because this needs to be fixed fast.

phstc:
@kyrylo on the missing HB thing was also because of `Honeybadger.context(some_hash)`, but I understand I can set key/value one by one.

kyrylo:
Can this work?

```rb
Thread.current[:airbrake_context] = { some: :hash }
```

phstc:
@kyrylo I guess I misread the PR then, I thought [this](https://github.com/airbrake/airbrake-ruby/pull/224/files#diff-c46d100454bdc9fc40c4e511b4ff283fR14) was whitelisting classes that could be used.

kyrylo:
Yeah, Array and Hash can also be attached. "unsafe" classes will be converted to strings. This test tests that: https://github.com/airbrake/airbrake-ruby/pull/224/files#diff-e0ec0dd90a467d38ced308e21683f5c4R75

phstc:
> converted to strings

@kyrylo so it will `to_s` a hash?

kyrylo:
Nope, Array and Hash will be attached as is. Other "unsafe" objects inside them will be converted to strings.

PR #226, opened 2017-05-23
notifier: don't run the specified block for ignored notices

kyrylo:


vmihailenco:
Does it work with the case when user ignores notice in the block?

kyrylo:
Good catch, it actually allows sending in that case. Working on a fix.

PR #233, opened 2017-06-18
Do a shallow copy of params hash in Notifier#build_notice

yanowitz:
Ran into a problem in a recent Airbrake upgrade where a call to Airbrake.notify
was (unexpectedly!) mutating the params hash.

This is one approach to fix.

At various points in the filter call chains of building up an Airbrake
notice, the code mutates the params hash that is passed in (adding
keys like :priority and safe_level). A simple dup means the calling
code no longer has to worry about that params hash mutating.


yanowitz:
forgot to do a bunch of the stuff for a decent PR. oops.

PR #234, opened 2017-06-18
Do a shallow copy of params hash in Notifier#build_notice

yanowitz:
Ran into a problem in a recent Airbrake upgrade where a call to Airbrake.notify
was (unexpectedly!) mutating the params hash.

This is one approach to fix.

At various points in the filter call chains of building up an Airbrake
notice, the code mutates the params hash that is passed in (adding
keys like :priority and safe_level). A simple dup means the calling
code no longer has to worry about that params hash mutating.

### DIFF
--- a/spec/notifier_spec.rb
+++ b/spec/notifier_spec.rb
@@ -525,7 +525,8 @@ RSpec.describe Airbrake::Notifier do
       it "prevents mutation of passed-in params hash" do
         params = { only_this_item: true }
         notice = @airbrake.build_notice(RuntimeError.new('bingo'), params)
-        expect(params.object_id).not_to eq(notice[:params].object_id)
+        notice[:params][:extra_item] = :not_in_original_params
+        expect(params).to eq(only_this_item: true)
       end
 
       it "returns the internal frames nevertheless" do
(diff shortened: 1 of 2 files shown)


yanowitz:
Hilarious... it let you approve it!

spec/notifier_spec.rb
```diff
@@ -522,6 +522,12 @@ def to_json(*)
     end
 
     context "given a non-exception with calculated internal frames only" do
+      it "prevents mutation of passed-in params hash" do
```
kyrylo: I don't think this test corresponds to the description. You check for the presence of `dup`, not if the object was mutated. Mutated objects carry the same object id. Thus, this test should be changed to something more sturdy:

```rb
params = { only_this_item: true }
notice = @airbrake.build_notice(RuntimeError.new('bingo'), params)
notice[:params][:extra_item] = true
expect(params).not_to include(:extra_item)
```

kyrylo:
That's a good catch 👍

spec/notifier_spec.rb
yanowitz: agreed, that's a much better test.

kyrylo:
Just released [v2.2.7](https://github.com/airbrake/airbrake-ruby/blob/master/CHANGELOG.md#v227-june-24-2017) with this fix.

PR #237, opened 2017-07-04
Implement `Airbrake#configured?`

kyrylo:
This helper method is intended to simplify the following check:

```ruby
if Airbrake[:default].is_a?(Airbrake::NilNotifier)
  # ...
end
```

With the new API it looks like this:

```ruby
if Airbrake.configured?
  # ...
end
```

### DIFF
--- a/lib/airbrake-ruby/notifier.rb
+++ b/lib/airbrake-ruby/notifier.rb
@@ -94,6 +94,12 @@ module Airbrake
       promise
     end
 
+    ##
+    # @macro see_public_api_method
+    def configured?
+      @config.valid?
+    end
+
     private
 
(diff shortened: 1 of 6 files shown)


lib/airbrake-ruby/notifier.rb
```diff
@@ -94,6 +94,12 @@ def create_deploy(deploy_params)
       promise
     end
 
+    ##
+    # @macro see_public_api_method
+    def configured?
+      true
```
vmihailenco: `@config.valid?`

kyrylo: Fixed.

PR #241, opened 2017-07-21
filter with raw exception instance

kuboon:
README shows following example to ignore ``StandardError``

```rb
  if notice[:errors].any? { |error| error[:type] == 'StandardError' }
    notice.ignore!
  end
end
```

But we rarely raise ``StandardError`` itself.
Instead, we raise the descendants of ``StandardError``.

Current implementation converts the exception to a string value before filtering.
It looses class hierarchy.

This patch allows filter to use exception instance to do like this:

```rb
  notice.ignore! if notice.exception.kind_of? StandardError
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -497,7 +497,7 @@ Instead, you can ignore notices based on some condition.
 
 ```ruby
 Airbrake.add_filter do |notice|
-  notice.ignore! if notice.exception.kind_of? StandardError
+  notice.ignore! if notice.exception.kind_of?(StandardError)
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -497,7 +497,7 @@ Instead, you can ignore notices based on some condition.
 
 ```ruby
 Airbrake.add_filter do |notice|
-  notice.ignore! if notice.exception.kind_of?(StandardError)
+  notice.ignore! if notice.stash[:exception].is_a?(StandardError)
 end
 ```
 
--- a/README.md
+++ b/README.md
@@ -497,9 +497,7 @@ Instead, you can ignore notices based on some condition.
 
 ```ruby
 Airbrake.add_filter do |notice|
-  if notice[:errors].any? { |error| error[:type] == 'StandardError' }
-    notice.ignore!
-  end
+  notice.ignore! if notice.exception.kind_of? StandardError
 end
 ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -497,9 +497,7 @@ Instead, you can ignore notices based on some condition.
 
 ```ruby
-  if notice[:errors].any? { |error| error[:type] == 'StandardError' }
-    notice.ignore!
-  end
+  notice.ignore! if notice.exception.kind_of? StandardError
```
kyrylo: Please use brackets around `.kind_of?(StandardError)`. We prefer this style in this project.

lib/airbrake-ruby/notice.rb
```diff
@@ -62,7 +62,7 @@ class Notice
     # @since v1.7.0
     # @return [Hash{Symbol=>Object}] the hash with arbitrary objects to be used
     #   in filters
-    attr_reader :stash
+    attr_reader :stash, :exception
```
kyrylo: We should probably move this on a separate line because the comment above describes `:stash`.

kuboon:
Failed on Rubocop. How can I avoid it?

kyrylo:
To fix the `Metrics/ClassLength`, please rebase on `master`. I made some refactoring, which should help with this issue.

lib/airbrake-ruby/notice.rb
```diff
@@ -64,6 +64,10 @@ class Notice
     #   in filters
     attr_reader :stash
 
+    ##
+    # @return [Exception] raw exception
+    attr_reader :exception
```
vmihailenco: what about putting this into `@stash.exception`?

kyrylo: I don't see why not. @kuboon, that will probably the preferred solution. Let me know what you think.

kuboon: no problem. I did not understand what is ``stash`` for.

kuboon:
I can't see the detail of Rubocop fail.
image

kyrylo:
You can run Rubocop locally.

```
Offenses:

lib/airbrake-ruby/notice.rb:77:16: C: Layout/SpaceInsideHashLiteralBraces: Space inside { missing. (https://github.com/bbatsov/ruby-style-guide#spaces-operators)
      @stash = {exception: exception}
               ^
lib/airbrake-ruby/notice.rb:77:37: C: Layout/SpaceInsideHashLiteralBraces: Space inside } missing. (https://github.com/bbatsov/ruby-style-guide#spaces-operators)
      @stash = {exception: exception}
                                    ^
```

kyrylo:
Great work, thanks 👍

PR #247, opened 2017-09-11
Add $PROGRAM_NAME to the Notice context

dasch:
We're running a lot of Ruby executables, but they all use the same stack (with different config). This makes it impossible to tell which executable experienced an error by looking at Airbrake.

By adding `$PROGRAM_NAME` to the context, I hope to solve that.

### DIFF
--- a/lib/airbrake-ruby/notice.rb
+++ b/lib/airbrake-ruby/notice.rb
@@ -67,6 +67,9 @@ module Airbrake
     def initialize(config, exception, params = {})
       @config = config
 
+      # The command used to execute the program.
+      params[:programName] = $PROGRAM_NAME
+
       @payload = {
         errors: NestedException.new(exception, @config.logger).as_json,
         context: context,


lib/airbrake-ruby/notice.rb
```diff
@@ -19,6 +19,7 @@ class Notice
     CONTEXT = {
       os: RUBY_PLATFORM,
       language: "#{RUBY_ENGINE}/#{RUBY_VERSION}".freeze,
+      program: $PROGRAM_NAME,
```
kyrylo: We can't add this property here because it's not compatible with our API. `params` or `environment` would suit more.

kyrylo:
It's a sensible idea to add that by default (you can always use a custom filter). Let's add it to `params`. To fix the current build failure, you can rebase on master (I just pushed a fix).

dasch:
@kyrylo I wasn't sure how to add it to `params`, and looking at the code, it felt at home in `environment` — is this okay?

kyrylo:
`environment` is okay, too. From what I can see it's still in `context` :)

1. We also need a test
2. We don't use snake case, so `program_name` instead of `programName`

dasch:
Ah, sorry about that. I see what you mean. I've added it to `params` instead.

kyrylo:
1. I can see that the build needs to be fixed
2. `program_name`, not `programName`

kyrylo:
Replaced by https://github.com/airbrake/airbrake-ruby/pull/251

PR #253, opened 2017-09-19
Add support for rate limiting by IP

kyrylo:


lib/airbrake-ruby/sync_sender.rb
```diff
@@ -14,6 +14,7 @@ class SyncSender
     # @param [Airbrake::Config] config
     def initialize(config)
       @config = config
+      @rate_limit_reset = Time.now
```
vmihailenco: This can be initialized to `0`

lib/airbrake-ruby/response.rb
```diff
@@ -52,5 +63,10 @@ def self.truncated_body(body)
       end
     end
     private_class_method :truncated_body
+
+    def self.rate_limit_delay(response)
+      Time.now + response['X-RateLimit-Delay'].to_i
+    end
```
vmihailenco: `now + delay` is time when rate limit is reset, not delay

vmihailenco: i.e. you need to rename method name

lib/airbrake-ruby/response.rb
```diff
@@ -52,5 +63,10 @@ def self.truncated_body(body)
       end
     end
     private_class_method :truncated_body
+
+    def self.rate_limit_delay(response)
+      Time.now + response['X-RateLimit-Delay'].to_i
```
vmihailenco: Does this work when header is missing? It would be nice if it does.

vmihailenco:
What about async sender?

lib/airbrake-ruby/response.rb
kyrylo: It works and equals to 0 because of type casting.

lib/airbrake-ruby/sync_sender.rb
kyrylo: I wish it could work, but it won't. We can't compare Time and Integers:

```hs
[4] pry(main)> Time.now > 0
ArgumentError: comparison of Time with 0 failed
from (pry):4:in `>'
[5] pry(main)> 0 < Time.now
ArgumentError: comparison of Fixnum with Time failed
from (pry):5:in `<'
```

lib/airbrake-ruby/response.rb
kyrylo: Renamed to `rate_limit_reset`

kyrylo:
> What about async sender?

It uses sync sender, so it gets this feature for free.

PTAL

PR #254, opened 2017-09-19
Update code doc blocks to reflect initialize method from KeysFilter

rgranadino:
Hello,

I was integrating Airbrake on a project which required some filtering to be done and ran into a small issue. The sample code in the code blocks didn't match what was implemented. It's not a huge difference but thought it'd be helpful if it was updated for new users.


kyrylo:
Nice catch 👍

PR #258, opened 2017-10-10
Implement code hunks support

kyrylo:
This feature is experimental. I am not documenting it in the README because
users shouldn't be using it yet.


lib/airbrake-ruby/backtrace.rb
```diff
@@ -188,6 +193,21 @@ def match_frame(regexp, stackframe)
 
         Patterns::GENERIC.match(stackframe)
       end
+
+      def read_code_hunk(logger, frame)
+        code_hunk = Airbrake::CodeHunk.new(frame[:file], frame[:line]).to_h
+        return unless code_hunk
+
+        unless code_hunk.key?(0)
+          frame[:code_hunk] = code_hunk
```
vmihailenco: Let's rename to just `code` for brevity? Anyway in JSON it should be `codeHunk`, not `code_hunk`

lib/airbrake-ruby/backtrace.rb
```diff
@@ -188,6 +193,21 @@ def match_frame(regexp, stackframe)
 
         Patterns::GENERIC.match(stackframe)
       end
+
+      def read_code_hunk(logger, frame)
+        code_hunk = Airbrake::CodeHunk.new(frame[:file], frame[:line]).to_h
+        return unless code_hunk
+
+        unless code_hunk.key?(0)
```
vmihailenco: I see 2 `unless` that check for `code_hunk`. I also don't understand what `code_hunk.key?(0)` does so it would be nice if you refactor this code or make it more obvious.

lib/airbrake-ruby/backtrace.rb
```diff
@@ -188,6 +193,21 @@ def match_frame(regexp, stackframe)
...
+          frame[:code_hunk] = code_hunk
+          return
+        end
+
+        logger.error(
+          "#{LOG_LABEL} error while reading code hunk `#{file}:#{line}'. " \
+          "Reason: #{code_hunk[0]}"
```
vmihailenco: Again, I don't understand why `code_hunk[0]` contains the reason.

lib/airbrake-ruby/code_hunk.rb
```diff
@@ -0,0 +1,51 @@
...
+  # Represents a small hunk of code consisting of a base line and a couple lines
+  # around
+  class CodeHunk
+    ##
+    # @return [Integer] the maximum length of a line
+    MAX_LINE_LEN = 200
+
+    ##
+    # @return [Integer] how many lines should be read around the base line
+    INTERVAL = 3
```
vmihailenco: `NLINES` would be a better name. Also I suggest to lower this to `2` so we have 2 lines before and 2 lines after.

spec/code_hunk_spec.rb
```diff
@@ -0,0 +1,89 @@
+require 'spec_helper'
+
+RSpec.describe Airbrake::CodeHunk do
+  describe "#to_h" do
+    context "when a file is empty" do
+      subject { described_class.new(fixture_path('empty_file.rb'), 1).to_h }
+
+      it { is_expected.to eql({}) }
+    end
```
vmihailenco: For empty file I expect to see `{1: ""}`

spec/code_hunk_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+    context "when a file doesn't exist" do
+      subject { described_class.new(fixture_path('banana.rb'), 1).to_h }
+
+      it { is_expected.to be_nil }
+    end
```
vmihailenco: Why not to throw an exception (return error)?

spec/code_hunk_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+            218 => '      end',
+            219 => '    end',
+            220 => '  end',
+            221 => 'end'
+          )
+        )
+      end
+    end
+
+    context "when a file has less lines than a code hunk requests" do
```
vmihailenco: Would be nice if you follow previous context and say something like `less than INTERVAL lines before and after`

spec/code_hunk_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+          eq(
+            1 => 'module Banana',
+            2 => '  attr_reader :bingo',
+            3 => 'end'
+          )
+        )
+      end
+    end
+
+    context "when a line location is in the middle of a file" do
```
vmihailenco: Looks like this should `has enough lines before and after`

lib/airbrake-ruby/code_hunk.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    def initialize(file, line, interval = INTERVAL)
+      @file = file
+      @line = line
+
+      @start_line = [line - interval, 1].max
+      @end_line = line + interval
+
+      @code_hash = {}
```
vmihailenco: @lines

lib/airbrake-ruby/backtrace.rb
kyrylo: 👍

lib/airbrake-ruby/backtrace.rb
kyrylo: `#to_h` returns a hash of `int => string`. A file can never have a zero line. I thought the existence of it would mean there's an error.

First `unless` is when a file is empty. Second `unless` is when there's an error that we need to log.

Does that make sense?

lib/airbrake-ruby/backtrace.rb
kyrylo: Files can never have line 0.

lib/airbrake-ruby/code_hunk.rb
kyrylo: 👍

lib/airbrake-ruby/code_hunk.rb
kyrylo: 👍

lib/airbrake-ruby/backtrace.rb
vmihailenco: It makes sense, but why not to throw an exception like any other Ruby code?

kyrylo: It would be unfortunate to miss a notice just because we couldn't read a line of code. I think simply ignoring that error and moving on would be the best solution.

spec/code_hunk_spec.rb
kyrylo: 👍

spec/code_hunk_spec.rb
kyrylo: What would it buy us?

spec/code_hunk_spec.rb
kyrylo: 👍

spec/code_hunk_spec.rb
kyrylo: 👍

kyrylo:
Pending discussion in https://github.com/airbrake/airbrake-ruby/pull/258#discussion_r143720560 and https://github.com/airbrake/airbrake-ruby/pull/258#discussion_r143722133 but the rest is fixed.

PTAL

lib/airbrake-ruby/backtrace.rb
vmihailenco: I don't suggest you throw that exception to users. You can catch and log it as usually.

spec/code_hunk_spec.rb
vmihailenco: AFAIK that is how errors are returned in Ruby.

kyrylo: If you have examples, I'd look at them. I don't think that's how it works. Either way, it's not a big deal because in real world backtraces would unlikely point to non existing files.

lib/airbrake-ruby/backtrace.rb
kyrylo: 👍

kyrylo:
PTAL

lib/airbrake-ruby/backtrace.rb
```diff
@@ -188,6 +193,13 @@ def match_frame(regexp, stackframe)
 
         Patterns::GENERIC.match(stackframe)
       end
+
+      def read_code_hunk(config, frame)
+        code_hunk = Airbrake::CodeHunk.new(config, frame[:file], frame[:line]).to_h
```
vmihailenco: Is there a good reason the method is called `to_h`? I see none.

lib/airbrake-ruby/backtrace.rb
```diff
@@ -176,10 +176,15 @@ def execjs_exception?(exception)
...
-        { file: match[:file],
+      def stack_frame(config, match)
+        frame = {
+          file: match[:file],
           line: (Integer(match[:line]) if match[:line]),
-          function: match[:function] }
+          function: match[:function]
+        }
+
+        read_code_hunk(config, frame) if config.code_hunks
```
vmihailenco: This is supposed to read something, but it returns nil.

lib/airbrake-ruby/code_hunk.rb
```diff
@@ -0,0 +1,57 @@
...
+        return
+      end
+
+      @lines = { 1 => '' } if @lines.empty?
+      @lines
+    end
+
+    private
+
+    def fetch_code
```
vmihailenco: This actually reads somethings, but it is called fetch and returns nil too.

lib/airbrake-ruby/code_hunk.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    ##
+    # @return [Hash{Integer=>String}, nil] code hunk around the base line. When
+    #   an error occurrs, returns a zero key Hash
+    def to_h
+      return @lines if @lines.any?
+      return unless File.exist?(@file)
+
+      begin
+        fetch_code
```
vmihailenco: What I see is that you check `lines`, then for some reason call `fetch_code` and then check `lines` again. It would be much more obvious if code was:

```
check @lines
set @lines
check @lines
```

lib/airbrake-ruby/code_hunk.rb
```diff
@@ -0,0 +1,57 @@
...
+      rescue StandardError => ex
+        @config.logger.error(
+          "#{self.class.name}##{__method__}: can't read code hunk for " \
+          "#{@file}:#{@line}: #{ex}"
+        )
+        return
+      end
+
+      @lines = { 1 => '' } if @lines.empty?
```
vmihailenco: Why this is here and not in `fetch_code`?

lib/airbrake-ruby/code_hunk.rb
```diff
@@ -0,0 +1,57 @@
...
+    ##
+    # @return [Hash{Integer=>String}, nil] code hunk around the base line. When
+    #   an error occurrs, returns a zero key Hash
+    def to_h
+      return @lines if @lines.any?
+      return unless File.exist?(@file)
+
+      begin
+        fetch_code
+      rescue StandardError => ex
```
vmihailenco: This is an example you asked for.

vmihailenco:
With this change notifier reads every file in backtrace for every notice which is a bit wasteful. I think we need to add a fixed size cache (say last 100 entries) in separate PR to make situation better.

lib/airbrake-ruby/backtrace.rb
kyrylo: `read`?

vmihailenco: I would expect `CodeHunk` to be shared and code look like this:

```
def constructor
  @code_hunk = Airbrake::CodeHunk.new(config)

def populate_code(frame)
  @code = @code_hunk.get(file, line)
  if @code
    frame[:code] = @code
```

lib/airbrake-ruby/code_hunk.rb
kyrylo: Technically, checking lines on line 29 is not needed. It's just prevention of reading the same file twice (for the same results). I removed it for now.

lib/airbrake-ruby/backtrace.rb
kyrylo: No big deal, I can init at the same place.

kyrylo:
> With this change notifier reads every file in backtrace for every notice which is a bit wasteful. I think we need to add a fixed size cache (say last 100 entries) in separate PR to make situation better.

I agree.

---

I simplified code.

PTAL

PR #259, opened 2017-10-11
code_hunk: cache file contents with simple cache

kyrylo:


lib/airbrake-ruby/code_hunk.rb
```diff
@@ -12,8 +12,9 @@ class CodeHunk
     # @return [Integer] how many lines should be read around the base line
     NLINES = 2
 
-    def initialize(config)
+    def initialize(config, cache)
       @config = config
+      @cache = cache
```
vmihailenco: Can you create cache here instead of passing it as arg?

kyrylo: It will be useless because it will be per frame, not per backtrace.

lib/airbrake-ruby/code_hunk.rb
```diff
@@ -44,5 +45,11 @@ def get(file, line)
       return { 1 => '' } if lines.empty?
       lines
     end
+
+    private
+
+    def get_from_cache(file)
+      @cache[file] ||= File.foreach(file)
```
vmihailenco: It is probably fine as is, but I was thinking about caching the result (code hunk given file&line) instead of whole file which can easily be 100kb * 100 cache entries = 10 mbs. I don't insist though.

kyrylo: We can always lower the limit of cache. 100 unique files per backtrace is very unrealistic (even 100 frames is hard to achieve).

lib/airbrake-ruby/code_hunk.rb
vmihailenco: Can we make CodeHunk and cache instances global?

kyrylo: Not sure what exactly you suggest, but that doesn't sound thread-safe friendly.

vmihailenco: I suggest that there is exactly one instance of `CodeHunk` in the notifier. And all code hunks are cached in single cache instance. Thread-safety can be achieved with mutexes where needed.

kyrylo:
@vmihailenco I pushed an update to use global cache (https://github.com/airbrake/airbrake-ruby/pull/259/commits/0b80da356ae375b0ce8e3a18107fdcc2f5ddc4eb). I realised that making it per notifier doesn't make sense because notifier accesses the file system, which is global to the ruby process. Two notifiers can access the same file and they would be read twice. With global cache 2 notifiers will share that file.

kyrylo:
PTAL

PR #264, opened 2017-10-16
code_hunk: handle `nil` lines & refactor to please Rubocop

kyrylo:
`nil` lines can happen when a backtrace contains file `config.ru` from
Rack. That frame doesn't carry line information.

Additionally, Rubocop was unhappy about the complexity of
`Airbrake::CodeHunk#get`, so I refactored it to make it simpler.

Tests were also leaking state because FileCache is global and we didn't nullify
it after every example. This is fixed now.

### DIFF
--- a/lib/airbrake-ruby/code_hunk.rb
+++ b/lib/airbrake-ruby/code_hunk.rb
@@ -22,26 +22,11 @@ module Airbrake
     # @return [Hash{Integer=>String}, nil] lines of code around the base line
     def get(file, line)
       return unless File.exist?(file)
+      return unless line
 
-      start_line = [line - NLINES, 1].max
-      end_line = line + NLINES
-      lines = {}
-
-      begin
-        get_from_cache(file).with_index(1) do |l, i|
-          next if i < start_line
-          break if i > end_line
-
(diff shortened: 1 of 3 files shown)


kyrylo:
Looks like the test is still flakey on Circle, but not locally. Needs more thought.

lib/airbrake-ruby/code_hunk.rb
```diff
@@ -22,33 +22,35 @@ def initialize(config)
...
-          lines[i] = l[0...MAX_LINE_LEN].rstrip
-        end
-      rescue StandardError => ex
-        @config.logger.error(
-          "#{self.class.name}##{__method__}: can't read code hunk for " \
-          "#{file}:#{line}: #{ex}\n#{ex.backtrace}"
-        )
-      end
-
+      populate_lines(lines, file, [line - NLINES, 1].max, line + NLINES)
```
vmihailenco: `lines = get_lines(file, [line - NLINES, 1].max, line + NLINES)`

kyrylo: 👍

PR #269, opened 2017-10-19
backtrace: define limiting logic for code hunks

kyrylo:
1. When `project_root` is configured, only frames within project have populated
   code hunks
2. When `project_root` isn't configured, only first 10 frames have populated
   code hunks


lib/airbrake-ruby/backtrace.rb
```diff
@@ -97,29 +97,18 @@ module Patterns
     end
 
     ##
+    # @return [Integer] how many first frames should include code hunks
+    CODE_FRAME_LIMIT = 10
```
vmihailenco: Ideally this should be auto-tunable option in truncator, e.g.
- try limit 64, check notice size
- try limit 32, check notice size
- etc

kyrylo:
It sounds interesting, however quite complex. Root directory will be set for the vast majority of projects (we even try to populate it by default), so this logic is plan B, to get at least some code hunks. I could agree if this was our main solution, but it’s a big time investment for something that’ll be used in 5% of projects or less.

PR #280, opened 2017-11-06
Enable truncation of frozen hashes

bengreville:
# Issue discovered in Truncator.rb
## Summary

Frozen hashes are unable to be truncated correctly by 
https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/truncator.rb#L103.

truncation error

## Description

Currently when truncating hashes for logging purposes, the original hash values get modified.

While immutability would be preferred, reusing the hash is likely here for keeping memory usage down. This is however creating an error when a frozen hash is passed in.

## Solution 👍 

This change resolves this by checking for a frozen hash, and duplicates it allowing for modification when required.

### DIFF
--- a/lib/airbrake-ruby/truncator.rb
+++ b/lib/airbrake-ruby/truncator.rb
@@ -101,6 +101,7 @@ module Airbrake
     end
 
     def truncate_hash(hash, seen)
+      hash = hash.dup if hash.frozen?
       hash.each_with_index do |(key, val), idx|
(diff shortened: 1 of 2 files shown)


bengreville:
Closing for now, while this works in isolation I believe breaking changes may occur for the implementation of `notice.rb` and the cache. More investigation will be required, which I'll get onto shortly

PR #283, opened 2017-11-08
Refactor Truncator to not mutate objects

### LINKED ISSUE #281: Frozen hashes breaking truncator functionality
# Issue discovered in Truncator.rb
## Summary

Frozen hashes are unable to be truncated correctly by 
https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/truncator.rb#L103.

truncation error

## Description

Currently when truncating hashes for logging purposes, the original hash values get modified.

While immutability would be preferred, reusing the hash is likely here for keeping memory usage down. This is however creating an error when a frozen hash is passed in.

## Potential Solution 👍 

This change resolves this by checking for a frozen hash, and duplicates it allowing for modification when required.

There is a pull request https://github.com/airbrake/airbrake-ruby/pull/280 (currently closed) outlining what a fix might look like, with related rspec tests included. 

**However there may be further mutations occurring upstream that I'm not aware of, which may make this a breaking change.**

Looking for input from the wider *airbrake-ruby developer community* to ensure the correct solution is found.

kyrylo:
Fixes #281 (Frozen hashes breaking truncator functionality)

### DIFF
--- a/lib/airbrake-ruby/truncator.rb
+++ b/lib/airbrake-ruby/truncator.rb
@@ -1,53 +1,34 @@
 module Airbrake
-  ##
   # This class is responsible for truncation of too big objects. Mainly, you
   # should use it for simple objects such as strings, hashes, & arrays.
   #
   # @api private
   # @since v1.0.0
   class Truncator
-    ##
     # @return [Hash] the options for +String#encode+
     ENCODING_OPTIONS = { invalid: :replace, undef: :replace }.freeze
 
-    ##
     # @return [String] the temporary encoding to be used when fixing invalid
     #   strings with +ENCODING_OPTIONS+
     TEMP_ENCODING = 'utf-16'.freeze
 
-    ##
     # @param [Integer] max_size maximum size of hashes, arrays and strings
     def initialize(max_size)
       @max_size = max_size
     end
 
-    ##
-    # Performs deep truncation of arrays, hashes and sets. Uses a
+    # Performs deep truncation of arrays, hashes, sets & strings. Uses a
     # placeholder for recursive objects (`[Circular]`).
     #
-    # @param [Hash,Array,Set] object The object to truncate
-    # @param [Hash] seen The hash that helps to detect recursion
-    # @return [void]
-    def truncate_object(object, seen = {})
-      return seen[object] if seen[object]
-
-      seen[object] = '[Circular]'.freeze
-      truncated =
-        if object.is_a?(Hash)
-          truncate_hash(object, seen)
-        elsif object.is_a?(Array)
-          truncate_array(object, seen)
-        elsif object.is_a?(Set)
-          truncate_set(object, seen)
-        else
-          raise Airbrake::Error,
-                "cannot truncate object: #{object} (#{object.class})"
-        end
-      seen[object] = truncated
+    # @param [Object] object The object to truncate
+    # @param [Set] seen The cache that helps to detect recursion
+    # @return [Object] truncated object
+    def truncate(object, seen = Set.new)
+      return '[Circular]'.freeze if seen.include?(object)
+      truncate_object(object, seen << object)
     end
 
-    ##
-    # Reduces maximum allowed size of the truncated object.
+    # Reduces maximum allowed size of hashes, arrays, sets & strings by half.
     # @return [Integer] current +max_size+ value
     def reduce_max_size
       @max_size /= 2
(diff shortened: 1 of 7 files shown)


lib/airbrake-ruby/truncator.rb
```diff
@@ -1,121 +1,96 @@
...
-      array.slice(0, @max_size).map! { |val| truncate(val, seen) }
+      array[0...@max_size].map { |elem| truncate(elem, seen) }.freeze
     end
 
     def truncate_set(set, seen)
-      set.keep_if.with_index { |_val, idx| idx < @max_size }.map! do |val|
-        truncate(val, seen)
-      end
+      Set.new(truncate_array(set.to_a, seen)).freeze
+    end
```
pond: This reads much more cleanly than the original code but I'm a bit worried about potential RAM usage. If this code is running, we must be looking at a Set with more than 10,000 items in it:

* https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/notice.rb#L80
* https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/notice.rb#L32

Above, we have the original Set with > 10,000 items, an Array copy with > 10,000 items that we're going to truncate as an Array, creating a 10,000 item Array, which we then convert to a 10,000 item Set.

I'm not very inventive :-) so thinking about how I tackled Hash (and given Set has more in common with Hash, really, than it does Array) we could do something like this:

```ruby
def truncate_set(set, seen)
  truncated_set = Set.new

  set.each do | val |
    truncated_set << truncate(val, seen)
    break if truncated_set.count >= @max_size
  end

  truncated_set.freeze
end
```

It's clearly verbose / clumsy and there may be room for improvement but it doesn't create any intermediate arrays.

lib/airbrake-ruby/truncator.rb
```diff
@@ -1,121 +1,96 @@
...
+        truncated_hash[key] = truncate(val, seen)
       end
+
+      truncated_hash.freeze
     end
 
     def truncate_array(array, seen)
-      array.slice(0, @max_size).map! { |val| truncate(val, seen) }
+      array[0...@max_size].map { |elem| truncate(elem, seen) }.freeze
     end
```
pond: This is a clean improvement accounting for frozen arrays nicely, but we end up creating two truncated arrays and can refine it to only have one. The `array[0...@max_size]` statement creates a copy of the original array over the requested slice, so that copy isn't frozen. But then we use `map` to avoid modifying the target, which builds another copy of the truncated array. The fix is super simple - we can modify that initially truncated array in-place - your improved code above works fine with `map!`:

```ruby
array[0...@max_size].map! { |elem| truncate(elem, seen) }.freeze
```

For more, see:

```ruby
array = [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 ].freeze
puts "Original: #{ array.object_id }"
short = array[0...6]
puts "Shortened: #{ short.object_id }"
mapped = short.map { | elem | elem }
puts "With 'map': #{ mapped.object_id } / #{ mapped.object_id == short.object_id }"
mapped = short.map! { | elem | elem }
puts "With 'map!': #{ mapped.object_id } / #{ mapped.object_id == short.object_id }"
```

...yields e.g.:

```
Original: 70230550309900
Shortened: 70230550310920
With 'map': 70230550313500 / false
With 'map!': 70230550310920 / true
```

lib/airbrake-ruby/truncator.rb
```diff
@@ -1,121 +1,96 @@
...
+    def truncate(object, seen = Set.new)
+      return '[Circular]'.freeze if seen.include?(object)
+      truncate_object(object, seen << object)
     end
 
-    ##
-    # Reduces maximum allowed size of the truncated object.
-    # @return [Integer] current +max_size+ value
+    # Reduces maximum allowed size of hashes, arrays and strings by half.
+    # @return [void]
```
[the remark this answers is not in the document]
kyrylo: Good catch, that was an oversight.

lib/airbrake-ruby/truncator.rb
kyrylo: Yeah, good catch, I think the only real addition this method needs is the `freeze` call.

lib/airbrake-ruby/truncator.rb
kyrylo: I like it, it's still simple 👍

kyrylo:
@pond addressed your feedback in https://github.com/airbrake/airbrake-ruby/pull/283/commits/692c07fc97256a190ea8019c9edc00f942ae9aa1.

PTAL

pond:
As a final note from our side, we tested for the problem in Hoodoo and worked around it in https://github.com/LoyaltyNZ/hoodoo/pull/229 released as https://rubygems.org/gems/hoodoo/versions/2.2.0 via https://github.com/LoyaltyNZ/hoodoo/pull/233. I'll track the progress of your patch and once it's merged will be able to revert the workaround and update Hoodoo's `gemspec` to require new enough versions of the dependencies.

PR #306, opened 2018-03-05
Also filter the hash in the array with KeysBlacklist

sawanoboly:
Hi, 

The current specification does not apply a filter if array contains parameter of notification contents. This PR also filters even if the filter target parameter is in array.


kyrylo:
Great PR 👍

kyrylo:
Released in v2.8.2.

Issue #312, opened 2018-03-21
TypeError: can't move to the enclosed thread group


meetme2meat:
Calling `Airbrake.notify` on Thread applications can result in a Thread been assign to an already enclosed ThreadGroup

The issue is very hard to replicate(but still we have 5-6 report of the above error in past 2 months) since it depends on the OS ability to context switch at wrong place resulting in a deadlock/race conditions on a shared object. 

The relevant code that been referred over here is `spawn_workers` inside the [async_sender.rb](https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/async_sender.rb#L95-L99)

Following is the backtrace reported by Airbrake.

```
TypeError: can't move to the enclosed thread group

org/jruby/RubyThreadGroup.java:87 in add
/gems/airbrake-ruby-1.1.0/lib/airbrake-ruby/async_sender.rb:93 in block in spawn_workers
org/jruby/RubyFixnum.java:296 in times
/gems/airbrake-ruby-1.1.0/lib/airbrake-ruby/async_sender.rb:93 in spawn_workers
/gems/airbrake-ruby-1.1.0/lib/airbrake-ruby/async_sender.rb:83 in has_workers?
/gems/airbrake-ruby-1.1.0/lib/airbrake-ruby/notifier.rb:130 in default_sender
/gems/airbrake-ruby-1.1.0/lib/airbrake-ruby/notifier.rb:45 in send_notice
/gems/airbrake-ruby-1.1.0/lib/airbrake-ruby/notifier.rb:45 in notify
/gems/airbrake-ruby-1.1.0/lib/airbrake-ruby.rb:282 in call_notifier
/gems/airbrake-ruby-1.1.0/lib/airbrake-ruby.rb:136 in notify
/srv/apps/policy_router/releases/20180212084654/lib/policy_router/response_handler.rb:52 in start
/srv/apps/policy_router/releases/20180212084654/lib/policy_router/request_handler.rb:7 in start
org/jruby/RubyKernel.java:1823 in public_send
/gems/celluloid-0.17.1.1/lib/celluloid/calls.rb:28 in dispatch
/gems/celluloid-0.17.1.1/lib/celluloid/call/sync.rb:16 in dispatch
/gems/celluloid-0.17.1.1/lib/celluloid/cell.rb:50 in block in dispatch
/gems/celluloid-0.17.1.1/lib/celluloid/cell.rb:76 in block in task
[...]

PR #313, opened 2018-03-21
Added a note on Airbrake.notify use.

meetme2meat:
https://github.com/airbrake/airbrake-ruby/issues/312: The Pull request is an attempt to notify the user of using an Airbrake.notify inside a Thread.


kyrylo:
Thanks for the PR. I believe this is a legit bug and it should be patched instead of documented. Please see https://github.com/airbrake/airbrake-ruby/pull/316.

PR #317, opened 2018-04-25
Implement basic API for merging context

### LINKED ISSUE #314: Possible to set context for possible error?
Is there a way to set some context that will be sent along by any downstream error notification?

Something like:
```
Airbrake.context(user_id: @user.id) do
  Foo.bar(@user)
end
```

The goal being that if anything within Foo.bar generates an Airbrake, the context of the user id will be included.  I don't see a way to do this with the current API, but perhaps I'm missing something.

kyrylo:
Fixes #314 (Possible to set context for possible error?)

This feature has been request many times already. While Airbrake
supports it, it's not obvious for users how to make it work. It's also
hard to advertise it since there's no simple method to call to explain
the API. Previously, I used to link people to
https://github.com/airbrake/airbrake/issues/696#issuecomment-366857391

I guess it is not very convenient to use that API. Therefore, the feature
doesn't use Threads and is much simpler to use. Users no longer need to mess
with threads and cleaning up context information is on us.

### DIFF
--- a/lib/airbrake-ruby/filters/context_filter.rb
+++ b/lib/airbrake-ruby/filters/context_filter.rb
@@ -1,0 +1,28 @@
+module Airbrake
+  module Filters
+    # Adds user context to the notice object. Clears the context after it's
+    # attached.
+    #
+    # @api private
+    # @since v2.9.0
+    class ContextFilter
+      # @return [Integer]
+      attr_reader :weight
+
+      def initialize(context)
+        @context = context
+        @weight = 119
+        @mutex = Mutex.new
+      end
+
+      def call(notice)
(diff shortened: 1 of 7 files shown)


lib/airbrake-ruby/filters/context_filter.rb
```diff
@@ -0,0 +1,20 @@
+module Airbrake
+  module Filters
+    # Adds user context to the notice object.
+    class ContextFilter
+      # @return [Integer]
+      attr_reader :weight
+
+      def initialize
+        @weight = 120
```
kyrylo: Note to myself. This may need tweaking.

PR #327, opened 2018-05-01
Introduce the 'versions' option

kyrylo:
We have been using `app_version` for tracking the user's app version. Since the
option expects a string there was no way to track any other
versions (dependencies is the most prominent example; version of Rails). We've
been using a hack where we prepend versions to the `app_version` value like this
'app/1.2.3 Rails/5.2.0`.

Those gloomy days are over with this new and shiny `versions` option. Now you
can attach any key and any value (please use names and versions respectively).
For example:

```ruby
config.versions = { 'bundler' => '1.2.3', 'Rails' => '5.2.0' }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,21 @@ Airbrake.configure do |c|
 end
 ```
 
+#### versions
+
+A hash of arbitrary versions that your application uses that you want to
+track (Rails, Linux kernel, API version, etc.). By default, it's empty.
+
+```ruby
+Airbrake.configure do |c|
+  c.versions = {
+    'Rails' => '5.2.0',
+    'bundler' => '1.16.1',
+    'LinuxKernel' => '4.4.0-97-generic'
+  }
+end
+```
+
 #### host
 
 By default, it is set to `airbrake.io`. A `host` is a web address containing a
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -222,6 +222,21 @@ Airbrake.configure do |c|
...
 ```
 
+#### versions
+
+A hash of arbitrary versions that your application uses that you want to
+track (Rails, Linux kernel, API version). By default, it's empty.
+
+```ruby
+Airbrake.configure do |c|
+  c.dependency_versions = {
```
vmihailenco: dependency_versions -> versions

README.md
```diff
@@ -222,6 +222,21 @@ Airbrake.configure do |c|
...
+    'Rails' => '5.2.0',
+    'bundler' => '1.16.1',
+    'Linux kernel' => '4.4.0-97-generic'
```
vmihailenco: It would be nice to remove spaces and use either `Linux_kernel` or `LinuxKernel` so it is easier to use in aggregations like `context.versions.LinuxKernel`. I am not even sure aggregations support spaces in keys.

README.md
kyrylo: Fixed.

README.md
kyrylo: 👍

kyrylo:
PTAL

PR #341, opened 2018-09-24
Attach git information for automatic deploys

kyrylo:
If you deploy your app with `.git` directory in the root, you can benefit from
Airbrake's automatic deployment feature. The notifier automatically detects the
presence of it and retrieves all required information to build a deploy for
you. No extra steps needed!

Note: this works only for certain accounts, so if you're reading it, you probably cannot use it just *yet*!


kyrylo:
@zefer I wasn't aware that this is experimental.
https://github.com/airbrake/gobrake implements it already and it's in public.

lib/airbrake-ruby/filters/git_last_checkout_filter.rb
```diff
@@ -0,0 +1,85 @@
...
+
+        parts = line.chomp.split("\t").first.split(' ')
+        if parts.size < 5
+          @logger.error(
+            "#{LOG_LABEL} Airbrake::#{self.class.name}: can't parse line: #{line}"
+          )
+          return
+        end
+
+        author = parts[2..4]
```
vmihailenco: I am not sure you can use `4` here, because this number will be different if user name contains space or not, e.g. "vmihailenco" or "Vladimir Mihailenco".

lib/airbrake-ruby/filters/git_last_checkout_filter.rb
```diff
@@ -0,0 +1,85 @@
...
+        @last_checkout = {
+          username: author[0..1].join(' '),
+          email: author.last[1..-2],
```
vmihailenco: What is there is no email?

spec/filters/git_last_checkout_filter_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      expect(notice[:context][:lastCheckout]).to be_nil
+    end
+  end
+
+  context "when .git directory exists" do
+    before { subject.call(notice) }
+
+    %i[username email revision time].each do |param|
+      it "attaches last checkouted #{param}" do
+        expect(notice[:context][:lastCheckout][param]).not_to be_empty
```
vmihailenco: It would be nice to see at least name and email here...

spec/filters/git_repository_filter.rb
```diff
@@ -0,0 +1,33 @@
...
+    it "doesn't attach anything to context/repository" do
+      subject.call(notice)
+      expect(notice[:context][:repository]).to be_nil
+    end
+  end
+
+  context "when .git directory exists" do
+    it "attaches context/repository" do
+      subject.call(notice)
+      expect(notice[:context][:repository]).not_to be_nil
```
vmihailenco: Can you include real repo name here? It should be the same every build.

spec/filters/git_last_checkout_filter_spec.rb
kyrylo: Unless we mock it, it will keep changing depending on who committed.

lib/airbrake-ruby/filters/git_last_checkout_filter.rb
kyrylo: When there's no email, then git will use hostname. I deleted my email and got: `<kyrylo@Kyrylos-MacBook-Pro.local>`

lib/airbrake-ruby/filters/git_last_checkout_filter.rb
kyrylo: Yeah, looks like you're right. It'll be broken.

kyrylo:
I've improved specs to be more specific and fixed possible bugs with username/author detection (by using negative indices).

PTAL

PR #375, opened 2019-01-17
filter_chain: add the ability to replace a default filter

kyrylo:
Sometimes users may not be happy with the default behaviour of our
filters (concrete scenario: https://github.com/airbrake/airbrake/issues/883). In
this case we want users to allow replacing a default filter they don't like.


kyrylo:
Closed in favour of 
https://github.com/airbrake/airbrake-ruby/pull/376.

PR #379, opened 2019-01-29
Add the ability to send SQL data

kyrylo:


lib/airbrake-ruby/notifier.rb
```diff
@@ -120,6 +121,21 @@ def notify_request(request_info)
       @route_sender.notify_request(request_info, promise)
     end
 
+    # @macro see_public_api_method
+    def notify_sql(sql_info)
```
vmihailenco: In gobrake we ended with:
- notifier.Routes.Notify(route_info), notifier.Routes.AddFilter
- notifier.Queries.Notify(query_info), notifier.Queries.AddFilter

Perhaps you can do something similar here...

kyrylo: Yeah, I also noticed that we can decouple this from the `Airbrake` module. I'll think about the design after this PR is merged, so it doesn't distract me from the ultimate goal of sending SQL.

README.md
```diff
@@ -780,6 +780,27 @@ promise.
...
+Airbrake.notify_query(
+  method: 'GET',
+  route: '/things/1',
+  query: 'SELECT * FROM foos',
+  start_time: Time.new - 200,
+  end_time: Time.new
+)
+```
+
+When [`config.route_stats = false`](#route_stats), it always returns a rejected
```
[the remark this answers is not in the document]
kyrylo: Either `query_stats` or `config.performance`. This really depends on how we want to structure this. E.g. do we want to allow configuring this individually or as a whole? I'd say for simplicity we want just 1 setting.

PR #394, opened 2019-02-05
Maintain UTC offset when truncating

sgray:
Previously, the time would be converted to UTC without properly adjusting for the given time's time zone.


kyrylo:
👍

PR #445, opened 2019-02-27
Return a rejected promise on #notify when Airbrake is not configured

kyrylo:
We want to be able to `add_filter`'s to Airbrake even when it's not
"configured". This resolves the headache with load order in the Rails
integration: add whenever you can but notify only when configured.

It also resolves the problem with `notify` methods that return `nil`. Users of
our API had to perform annoying `nil` checking, now they don't need this and can
fully rely on promises.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Airbrake Ruby Changelog
 
 ### master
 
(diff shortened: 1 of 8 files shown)


kyrylo:
Replaced by https://github.com/airbrake/airbrake-ruby/pull/450.

PR #451, opened 2019-02-28
Delete NilNotifiers and rely on notifier config checks

kyrylo:
When a notifier is not configured, we can `add_filter` but `notify` won't have
effect. This accompanies https://github.com/airbrake/airbrake-ruby/pull/445.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ Airbrake Ruby Changelog
 
 ### master
 
+* `add_filter` & `add_performance_filter` add filters even when Airbrake is not
+  configured ([#445](https://github.com/airbrake/airbrake-ruby/pull/445),
+  [#451](https://github.com/airbrake/airbrake-ruby/pull/451))
+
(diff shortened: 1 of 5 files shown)


ivanovv:
I believe this PR brought back #80

ivanovv:
C'mon, guys, this is ridiculous, I'm trying to revert to an older version, and the only alternatives I have are
* `airbrake 8.1.4` that uses `airbrake-ruby 3.2.6` and throws ` warning: Airbrake#notifiers is deprecated...` all over the place (and that's pretty annoying to say the least)
* `airbrake 8.2.1` that uses `airbrake-ruby 4.1.0` and crashes on CircleCi with `Airbrake::Error: :project_id is required`
* something older than that, but I then I have to quit performance beta program

ivanovv:
I know a workaround for this, but I don't understand why I have to spend time searching for workarounds for something that we pay money for

kyrylo:
Can you please open a new issue and share your config?

ivanovv:
https://github.com/airbrake/airbrake/issues/931

PR #493, opened 2019-07-18
Update README.md

krisdigital:
Named notifiers do not exist anymore.


kyrylo:
Good catch, thank you!

PR #507, opened 2019-10-04
filters/sql: ignore more irrelevant queries from Postgres

kyrylo:


kyrylo:
These queries are executed not because of user app's code (e.g. a model JOINed some other model because I wrote it like that in a controller), but because postgres needs them for its own internal stuff. Displaying these queries is irrelevant to 95% of users and it is just noise.

If other database engines include similar queries, we will be ignoring them in a similar fashion.

zefer:
Yeah, it makes sense, but on the flip-side, maybe we want to see all the queries, and it could be hard to do this for all engines.

kyrylo:
I was judging from a viewpoint of a typical Rails developer. I doubt they will need it. Even super experienced users will likely not benefit from knowing that an internal postgres request was completed in a timely fashion (which is probably how it'll be). We've already shipped this filter with some rules, so that ship has sailed 😁

It won't be hard to maintain a blacklist, especially if users will be reporting queries that they don't find useful.

zefer:
No ship has sailed at this early stage.

Anyway, I don't mean to stop you, it's just something to consider and we can change our approach later if needed. A couple of other options...

- don't filter queries so all is monitored and visible to users
- make it a notifier config option, so users can override if necessary
- filter on the UI: either lower the contrast, hide them, or some other approach to show they are not 'app-code'

kyrylo:
Thanks! Yeah, I agree, that can definitely work if we discover that people really want those service queries.

PR #513, opened 2019-12-05
Add support for `Airbrake.notify_queue`

kyrylo:


README.md
```diff
@@ -846,6 +846,46 @@ a rejected promise.
 When [`config.performance_stats = true`](#performance_stats), then it aggregates
 statistics and sends as a batch every 15 seconds.
 
+#### Airbrake.notify_queue
+
+Sends queue (job) statistics to Airbrake. Supports groups (similar to
```
kyrylo: @vmihailenco does this make sense? I am wondering, wouldn't `error_count` always be 1? A Worker would call this method and pass either `error_count: 0` when successful or `error_count: 1` if failed.

vmihailenco: For a single job - yes, error count can either be 0 or 1. So it does not make much sense here.

API uses it because there are multiple jobs and number of errors make more sense there.

kyrylo: How should the client API look like?

When `notify_queue` is called, it represents only one job. Should the error count be internal? That is, the user shouldn’t be able to specify it, the library will insert 1 or 0 based on the method invocation.

The more times you call the method with the same queue name, that many times normal count is bumped. Is that true?

vmihailenco: Just ignore that field for now.

PR #537, opened 2020-01-20
Fix crashing `add_performance_filter` example

ghiculescu:
Since v10, `airbrake` is able to send performance info for `ActiveJob` jobs: https://github.com/airbrake/airbrake/commit/8bff73b59f7c715d13a0108119c662cce65f005c

The payload is an object of type `Request::Queue`, which doesn't respond to `route`. Therefore, this PR fixes the sample code to ensure you check if the `route` method exists before calling it.


kyrylo:
Good catch! I merged https://github.com/airbrake/airbrake-ruby/pull/539 instead of this PR because we don't want to pollute filters with `respond_to?` checks. The original idea is that all "resources" respond to `route` and Queue must comply. Thanks for raising this issue!

PR #538, opened 2020-01-21
Making sure 2.7.0 final works

BrianHawley:


BrianHawley:
OK, so kyrylo/rbtree3#6 definitely affects things here as well. With the fix to that applied locally, I was able to fix a few other warnings. So far so good.

However, it looks like there is a bug in Ruby's Struct class in 2.7 where it gives warnings if a derived class implements its `initialize` method using keyword parameters. I'm not on Ruby's issue list yet, so I don't know if this has been reported. Going to see if there's a decent workaround.

.circleci/config.yml
```diff
@@ -92,7 +92,7 @@ jobs:
       - <<: *unit
   "ruby-2.7":
     docker:
-      - image: circleci/ruby:2.7.0-preview1
+      - image: circleci/ruby:2.7.0
```
kyrylo:
```suggestion
      - image: circleci/ruby:2.7
```

kyrylo:
Looking good!
Let's get rbtree3 released first, then I'll think about other warnings and specifically the `initialize` issue.

.circleci/config.yml
BrianHawley: Turns out that circleci doesn't have a 2.7 gag, only a 2.7.0 tag.

kyrylo: Wow, that's unexpected. Fair enough then *shrug*

BrianHawley:
I would have fixed the initialize issue already, but I had a party to go to. Will make another PR tomorrow.

PR #540, opened 2020-01-21
Fix all Ruby 2.7 warnings

BrianHawley:
- Allow rake 13.x on Ruby 2.2+.
- The whole safe_level feature is a deprecated noop in 2.7+.
- Fixed all keyword parameter warnings.
- KeywordStruct wrapper to allow Struct with keyword parameters on
  Ruby 2.7, or faking the behavior on other Ruby versions/platforms.

### DIFF
--- a/lib/airbrake-ruby.rb
+++ b/lib/airbrake-ruby.rb
@@ -84,6 +84,13 @@ module Airbrake
   #  special cases where we need to work around older implementations
   JRUBY = (RUBY_ENGINE == 'jruby')
 
+  # @return [Boolean] true if this Ruby supports safe levels and tainting,
+  #  to guard against using deprecated or unsupported features.
+  HAS_SAFE_LEVEL = (
+    RUBY_ENGINE == 'ruby' &&
+    Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.7')
+  )
+
   class << self
     # @since v4.2.3
     # @api private
--- a/lib/airbrake-ruby.rb
+++ b/lib/airbrake-ruby.rb
@@ -371,7 +371,7 @@ module Airbrake
     # @since v3.0.0
     # @see Airbrake::PerformanceNotifier#notify
     def notify_request(request_info, stash = {})
-      request = Request.new(request_info)
+      request = Request.new(**request_info)
       request.stash.merge!(stash)
       performance_notifier.notify(request)
     end
@@ -381,7 +381,7 @@ module Airbrake
     # @since v4.10.0
     # @see .notify_request
     def notify_request_sync(request_info, stash = {})
-      request = Request.new(request_info)
+      request = Request.new(**request_info)
       request.stash.merge!(stash)
       performance_notifier.notify_sync(request)
     end
@@ -414,7 +414,7 @@ module Airbrake
     # @since v3.2.0
     # @see Airbrake::PerformanceNotifier#notify
     def notify_query(query_info, stash = {})
-      query = Query.new(query_info)
+      query = Query.new(**query_info)
       query.stash.merge!(stash)
       performance_notifier.notify(query)
     end
@@ -425,7 +425,7 @@ module Airbrake
     # @since v4.10.0
     # @see .notify_query
     def notify_query_sync(query_info, stash = {})
-      query = Query.new(query_info)
+      query = Query.new(**query_info)
       query.stash.merge!(stash)
       performance_notifier.notify_sync(query)
     end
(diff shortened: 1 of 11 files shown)


kyrylo:
Not trying to nitpick, but it would be nice to split the changes you made into separate commits for the benefit of good git history. At the moment we combine multiple different changes as one commit, which is not good.

lib/airbrake-ruby.rb
```diff
@@ -84,6 +85,10 @@ module Airbrake
   #  special cases where we need to work around older implementations
   JRUBY = (RUBY_ENGINE == 'jruby')
 
+  # @return [Boolean] true if current Ruby is MRI < 2.7. The result is used
+  #  to guard against using deprecated features which will be removed in MRI 3.
+  HAS_SAFE_LEVEL = (RUBY_ENGINE == 'ruby' && RUBY_VERSION < '2.7')
```
kyrylo: We use `Gem::Version.new(RUBY_VERSION)` for comparisons. This is because it can handle corner cases that simple string comparison cannot.

kyrylo:
Thanks for the PR! I'll review this properly tomorrow.

BrianHawley:
I'll break the changes up into per-topic commits and do that version change.

lib/airbrake-ruby.rb
```diff
@@ -84,6 +84,14 @@ module Airbrake
   #  special cases where we need to work around older implementations
   JRUBY = (RUBY_ENGINE == 'jruby')
 
+  # Defined here so they can be used multiple times.
+  ruby_version_27 = Gem::Version.new('2.7')
+  ruby_version = Gem::Version.new(RUBY_VERSION)
+
+  # @return [Boolean] true if this Ruby supports safe levels and tainting,
+  #  to guard against using deprecated or unsupported features.
+  HAS_SAFE_LEVEL = (RUBY_ENGINE == 'ruby' && ruby_version < ruby_version_27)
```
kyrylo: I would just inline the variables since they don't make the code any clearer (and the comment on line 87 is redundant).

```suggestion
  HAS_SAFE_LEVEL = (
    RUBY_ENGINE == 'ruby' && 
    Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.7')
  )
```

lib/airbrake-ruby.rb
```diff
@@ -11,6 +11,7 @@
 require 'airbrake-ruby/stashable'
 require 'airbrake-ruby/mergeable'
 require 'airbrake-ruby/grouppable'
+require 'airbrake-ruby/keyword_struct'
```
kyrylo: While the implementation looks correct to me and it probably allows smooth transition to the new APIs, I'm not a big fan of the approach. We use metaprogramming to mimic the old behaviour but the code looks rather complex. People will need to learn about `KeywordStruct` and I think we could avoid it by refactoring the "model" structs into simple classes. It'll be a little bit more verbose but we wouldn't need to introduce any new concepts/classes. For example:

```diff
diff --git a/lib/airbrake-ruby/queue.rb b/lib/airbrake-ruby/queue.rb
index 275d972..e27b8a6 100644
--- a/lib/airbrake-ruby/queue.rb
+++ b/lib/airbrake-ruby/queue.rb
@@ -5,24 +5,31 @@ module Airbrake
   # @api public
   # @since v4.9.0
   # rubocop:disable Metrics/BlockLength, Metrics/ParameterLists
-  Queue = Struct.new(
-    :queue, :error_count, :groups, :start_time, :end_time, :timing, :time
-  ) do
+  class Queue
     include HashKeyable
     include Ignorable
     include Stashable

+    attr_reader :queue, :error_count, :groups, :start_time, :end_time, :timing,
+                :time
+
     def initialize(
-      queue:,
-      error_count:,
-      groups: {},
-      start_time: Time.now,
-      end_time: start_time + 1,
-      timing: nil,
-      time: Time.now
-    )
+          queue:,
+          error_count:,
+          groups: {},
+          start_time: Time.now,
+          end_time: start_time + 1,
+          timing: nil,
+          time: Time.now
+        )
       @time_utc = TimeTruncate.utc_truncate_minutes(time)
-      super(queue, error_count, groups, start_time, end_time, timing, time)
+      @queue = queue
+      @error_count = error_count
+      @groups = groups
+      @start_time = start_time
+      @end_time = end_time
+      @timing = timing
+      @time = time
     end

     def destination
```

Let me know what you think.

BrianHawley: That's OK with me. Your call.

I'm going to save KeywordStruct for future reference though in other projects (not here).

lib/airbrake-ruby.rb
BrianHawley: OK.

BrianHawley:
cc146f1 is relates to ruby warnings and stack overflows in the airbrake gem, to be fixed by airbrake/airbrake/1063. It's a side effect of having the airbrake config managed in this gem, rather than that one. I can split it out into a separate PR if you like.

BrianHawley:
@kyrylo will this do?

kyrylo:
Perfect, @BrianHawley, thanks a lot!

PR #542, opened 2020-01-22
Configuration fixes for airbrake/airbrake#1062

BrianHawley:
Airbrake::Rack::Instrumentable#airbrake_capture_timing does method
wrapping similar to alias_method_chain, but this style of wrapping
will cause stack overflows on methods that have been overriden
with Module#prepend. However, using prepend style wrapping will
stack overflow on methods built with alias_method_chain or the
equivalent alias_method calls.

Some people have chained codebases, like those based on Rails < 5.
Rails 5+ codebases are based on prepend. For either case, there
will be random methods that are implemted in the other style, so
even if we set a default we need to have the option of using the
other style. We need to support both wrapping methods, and we
need a way for the user to declare which wrapping method is used
as the default for their project. It can't be autodetected.

Method wrapping also needs to do argument forwarding, and the
syntax and semantics for that changed in Ruby 2.7, effectively.
The real change is coming in Ruby 3 but 2.7 adds many warning
messages for the transition.

Here, we just add the config setting for the default method
wrapping style, and a flag constant to indicate whether the
new argument forwarding style and syntax is supported. As new
Ruby platforms support the new syntax, the flag constant can
be adjusted to match.

The rest will be done in the airbrake gem.

### DIFF
--- a/lib/airbrake-ruby.rb
+++ b/lib/airbrake-ruby.rb
@@ -91,6 +91,12 @@ module Airbrake
     Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.7')
   )
 
+  # @return [Boolean] true if this Ruby supports ... argument forwarding.
+  HAS_ARGUMENT_FORWARDING = (
+    RUBY_ENGINE == 'ruby' &&
+    Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.7')
+  )
+
   class << self
     # @since v4.2.3
     # @api private
--- a/lib/airbrake-ruby/config.rb
+++ b/lib/airbrake-ruby/config.rb
@@ -107,6 +107,11 @@ module Airbrake
     # @since v4.12.0
     attr_accessor :job_stats
 
+    # @return [Symbol] technique used by default to wrap methods:
+    #   :chain for alias_method_chain style or :prepend for Module#prepend.
+    # @api public
+    attr_reader :default_wrapping_style
+
     class << self
       # @return [Config]
       attr_writer :instance
@@ -119,6 +124,7 @@ module Airbrake
 
     # @param [Hash{Symbol=>Object}] user_config the hash to be used to build the
     #   config
+    # rubocop:disable Metrics/AbcSize
     def initialize(user_config = {})
       self.proxy = {}
       self.queue_size = 100
@@ -147,8 +153,11 @@ module Airbrake
       self.query_stats = true
       self.job_stats = true
 
+      self.default_wrapping_style = :chain
+
       merge(user_config)
     end
+    # rubocop:enable Metrics/AbcSize
 
     # The full URL to the Airbrake Notice API. Based on the +:host+ option.
     # @return [URI] the endpoint address
@@ -165,6 +174,15 @@ module Airbrake
     # @return [Logger] the logger
     def logger=(logger)
       @logger = logger || @logger
+    end
+
+    # Sets the default method wrapping style. Must be :chain or :prepend.
+    def default_wrapping_style=(style)
+      style = style.to_sym if style.respond_to?('to_sym')
+      unless %i[chain prepend].include?(style)
+        raise ArgumentError, "default_wrapping_style must be :chain or :prepend"
+      end
+      @default_wrapping_style = style
     end
 
     # Merges the given +config_hash+ with itself.
--- a/spec/config_spec.rb
+++ b/spec/config_spec.rb
@@ -163,4 +163,43 @@ RSpec.describe Airbrake::Config do
       end
     end
   end
+
+  describe "#default_wrapping_style" do
+    let(:message) { "default_wrapping_style must be :chain or :prepend" }
+    it "is :chain by default" do
+      expect(subject.default_wrapping_style).to eq :chain
+    end
+
+    %i[chain prepend].each do |value|
+      it "allows :#{value}" do
+        subject.default_wrapping_style = value
+        expect(subject.default_wrapping_style).to eq value
+      end
+
+      it "allows '#{value}'" do
+        subject.default_wrapping_style = value.to_s
+        expect(subject.default_wrapping_style).to eq value
+      end
+    end
+
+    it "doesn't allow other symbols" do
+      expect { subject.default_wrapping_style = :foo }
+        .to raise_error(ArgumentError, message)
+    end
+
+    it "doesn't allow other strings" do
+      expect { subject.default_wrapping_style = 'bar' }
+        .to raise_error(ArgumentError, message)
+    end
+
+    it "doesn't allow other values" do
+      expect { subject.default_wrapping_style = 1 }
+        .to raise_error(ArgumentError, message)
+    end
+
+    it "doesn't allow nil" do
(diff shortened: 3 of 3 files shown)


BrianHawley:
Whoops, wrong branch name. Too late to change now.

BrianHawley:
This will need to be rebased after #540 is merged.

lib/airbrake-ruby/config.rb
```diff
@@ -167,6 +176,15 @@ def logger=(logger)
       @logger = logger || @logger
     end
 
+    # Sets the default method wrapping style. Must be :chain or :prepend.
+    def default_wrapping_style=(style)
+      style = style.to_sym if style.respond_to?('to_sym')
```
kyrylo: Why do we need to check if `style` responds to `to_sym`?
The expected argument is a String or a Symbol. They both seem to support `to_sym` since ancient times.
If we pass any other object, it'll fail with a `NoMethodError` and that's okay.

lib/airbrake-ruby/config.rb
```diff
@@ -167,6 +176,15 @@ def logger=(logger)
...
+      unless %i[chain prepend].include?(style)
+        raise ArgumentError "default_wrapping_style must be :chain or :prepend"
```
kyrylo:
```suggestion
        raise ArgumentError, "default_wrapping_style must be :chain or :prepend"
```

lib/airbrake-ruby/config.rb
```diff
@@ -167,6 +176,15 @@ def logger=(logger)
       @logger = logger || @logger
     end
 
+    # Sets the default method wrapping style. Must be :chain or :prepend.
+    def default_wrapping_style=(style)
```
kyrylo: We probably want to test this method (line 182).

lib/airbrake-ruby.rb
```diff
@@ -84,6 +84,12 @@ module Airbrake
   #  special cases where we need to work around older implementations
   JRUBY = (RUBY_ENGINE == 'jruby')
 
+  # @return [Boolean] true if this Ruby supports ... argument forwarding.
+  HAS_ARGUMENT_FORWARDING = (
```
kyrylo: This constant now looks exactly the same as `Airbrake::HAS_SAFE_LEVEL`.
We could give it another name, so it's more general and we could use it everywhere but I struggle to suggest anything decent. 🤷‍♂️

kyrylo:
> Airbrake::Rack::Instrumentable#airbrake_capture_timing does method
wrapping similar to alias_method_chain, but this style of wrapping
will cause stack overflows on methods that have been overriden
with Module#prepend. 

Could you show an example of this? I coined this example but I cannot see any overflows:

```rb
require 'airbrake'
require 'airbrake/rack'

module M
  def m
    :module
  end
end

class A
  prepend M
  extend Airbrake::Rack::Instrumentable

  def m
    :class
  end
  airbrake_capture_timing :m
end

puts A.new.m #=> "module"
```

lib/airbrake-ruby/config.rb
BrianHawley: I prefer to raise better errors. Unhandled NoMethodError is almost always an indication of a bug or a typo in your code, so if you have your code raise NoMethodError as part of your official API then your tests won't be able to tell the difference between that NoMethodError and one that was raised because you mistyped something in your code.

lib/airbrake-ruby/config.rb
BrianHawley: That's what I get for not writing a test for that line.

lib/airbrake-ruby.rb
BrianHawley: HAS_SAFE_LEVEL is true if MRI and version < 2.7, and probably never any other Ruby platforms.

HAS_ARGUMENT_FORWARDING is true for MRI and >= 2.7, and eventually might be true for JRuby and TruffleRuby once they start doing 2.7+ compatibility.

BrianHawley:
I'll try to write an example of code that will cause a stack overflow, for both the chain and prepend cases.

lib/airbrake-ruby.rb
BrianHawley: I had trouble coming up with a name too. HAS_ARGUMENT_FORWARDING is at least topical; it refers to the `def a(...); b(...); end` argument forwarding syntax.

lib/airbrake-ruby/config.rb
BrianHawley: I verified this in my tests. I added the tests and checked for ArgumentError, but didn't make the comma change below. It raised a NoMethodError instead, indicating a bug. If I had been expecting a NoMethodError I wouldn't have caught the bug. It's almost never good practice to raise NoMethodError on purpose.

BrianHawley:
The stack overflow test will need to go in https://github.com/airbrake/airbrake/pull/1066

lib/airbrake-ruby/config.rb
BrianHawley: Added tests.

lib/airbrake-ruby.rb
kyrylo: Oh yeah, my bad for not noticing they are different. Let's just disregard this comment then.

As for TruffleRuby, we might not even support it. I've never tried running tests there.

`HAS_ARGUMENT_FORWARDING` sounds like a good name to me, and it's easy to understand.

kyrylo:
>  I've seen it actually happen in other cases where I've written similar code in the past, and during Rails upgrades. It's the reason that they removed alias_method_chain in Rails 5.

I managed to reproduce it with the following snippet:

```rb
require 'airbrake'
require 'airbrake/rack'

module M
  def foo
    p 'M#foo'
    super
  end
end

class C
  extend Airbrake::Rack::Instrumentable
  def foo
    p 'C#foo'
  end
  prepend M
  airbrake_capture_timing :foo
end

C.new.foo
```

spec/config_spec.rb
```diff
@@ -163,4 +163,43 @@
...
+        expect(subject.default_wrapping_style).to eq value
+      end
+
+      it "allows '#{value}'" do
+        subject.default_wrapping_style = value.to_s
+        expect(subject.default_wrapping_style).to eq value
+      end
+    end
+
+    it "doesn't allow other symbols" do
```
kyrylo: We can probably shrink this with a loop (similarly to line 173):

```rb
[:foo, 'bar', 1, nil].each do |value|
  it "doesn't allow #{value.inspect}" do
    subject.default_wrapping_style = value
    expect { subject.default_wrapping_style = 'bar' }
      .to raise_error(ArgumentError, message)
  end
end
```

lib/airbrake-ruby/config.rb
kyrylo: FWIW we wouldn't raise a `NoMethodError` on purpose. We would just not perform any "typechecking". That said, what you said sounds reasonable to me, so no objections (especially since this will be called only once, on process initialisation).

kyrylo:
After a fruitful discussion in https://github.com/airbrake/airbrake/pull/1066 we realised that this PR is no longer necessary.

PR #579, opened 2020-06-16
Changes the terminology from whitelist_keys to allowlist_keys and blacklist_keys to blocklist_keys

ryanatwork:
* This PR changes the term `whitelist_keys` to `allowlist_keys` and `blacklist_keys` to `blocklist_keys`.
* Also changes the filter classes to `Airbrake::Filters::KeysBlocklist` and `Airbrake::Filters::KeysAllowlist`
* Since this is a breaking change I made it backwards compatible with the existing keys and added the deprecated note in the YARD docs.
* There is logic added though if both keys are set one will supersede the other. An example of this is if the config has both a `c.whitelist_keys` and `c.allowlist_keys` configured the configuration will take `allowlist_keys`.

Background:
* The technology industry has began moving away from terms such as master/slave and white list /blacklist.
* A recent tweet about this https://twitter.com/leahculver/status/1269109776983547904


destari:
Thanks!  Good eye, we'll take a look!

kyrylo:
Hey @ryanatwork, thanks for the PR.

I published an alternative version https://github.com/airbrake/airbrake-ruby/pull/580

The differences:

* we print a warning when an old option is used
* the implementation is much simpler
* the tests are simpler as well
* cleaner git history

We could've gotten there through the review process but I wanted to merge this rather sooner than later. Thanks again!

ryanatwork:
Thanks @destari and @kyrylo for the quick response on this! Very much appreciated.

PR #611, opened 2020-08-13
Make it possible to configure remote config host

kyrylo:
At the moment there's a bug with remote configuration. A user would install the
new notifier but the notifier would use our staging S3 bucket to fetch
configs. Instead, we should hardcode the bucket path to production because
that's the goal. A helper option called `remote_config_host` was added. With
help of it can easily test the functionality.


kyrylo:
It's not possible disable remote configuration. Could you include full backtrace (or at least stack frames from airbrake-ruby), so that I can see where the error is coming from?

PR #619, opened 2020-09-27
Add configuration to disable remote settings

derSascha:
On internal systems, external sites might be blocked and i guess many people are concerned about leaking information of their systems without asking them before... Another very bad idea is to [write into gem directories](https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/remote_settings.rb#L142).

This pull request adds a new configuration option to enable/disable the remote configuration feature and adds some information about it to the readme. One of the tests does not like my e-mail address, the pull request includes a fix to support subdomains in the test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,19 @@ Airbrake.notify('App crashed!', {
 #     credit_card: '[Filtered]' }
 ```
 
+#### remote_config and remote_config_host
+
+Airbrake Ruby can fetch its configuration from a remote site and updates them in
+regular intervals. By default, this feature is disabled and uses the host
+`https://v1-production-notifier-configs.s3.amazonaws.com`.
+
+```ruby
+Airbrake.configure do |c|
+  c.remote_config = true
+  c.remote_config_host = 'https://some-configuration-bucket.s3.amazonaws.com'
+end
+```
+
 ##### Using Procs to delay filters configuration
 
 If you cannot inline your keys (for example, they should be loaded from external
--- a/lib/airbrake-ruby/config.rb
+++ b/lib/airbrake-ruby/config.rb
@@ -124,6 +124,12 @@ module Airbrake
     # @since 5.0.0
     attr_accessor :error_notifications
 
+    # @return [Boolean] true if the library should fetch remote configurations,
+    #   false otherwise
+    # @api public
+    # @since v5.0.3
+    attr_accessor :remote_config
+
     # @return [String] the host such as which should be used for fetching remote
     #   configuration options (example: "https://bucket-name.s3.amazonaws.com")
     attr_accessor :remote_config_host
@@ -140,7 +146,7 @@ module Airbrake
 
     # @param [Hash{Symbol=>Object}] user_config the hash to be used to build the
     #   config
-    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
     def initialize(user_config = {})
       self.proxy = {}
       self.queue_size = 100
@@ -151,6 +157,7 @@ module Airbrake
       self.project_key = user_config[:project_key]
       self.error_host = 'https://api.airbrake.io'
       self.apm_host = 'https://api.airbrake.io'
+      self.remote_config = false
       self.remote_config_host = 'https://v1-production-notifier-configs.s3.amazonaws.com'
 
       self.ignore_environments = []
@@ -174,7 +181,7 @@ module Airbrake
 
       merge(user_config)
     end
-    # rubocop:enable Metrics/AbcSize
+    # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
 
     # The full URL to the Airbrake Notice API. Based on the +:error_host+ option.
     # @return [URI] the endpoint address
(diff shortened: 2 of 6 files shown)


kyrylo:
Thanks for the PR! I will consult my team about the issues that you raised. Specifically, "leaking information of their systems without asking them before".

> Another very bad idea is to write into gem directories.

Could you elaborate more about that? What makes it "very bad"?

> One of the tests does not like my e-mail address, the pull request includes a fix to support subdomains in the test.

If you open a new PR just with that change, I will merge it immediately. The rest of the code will have to wait for my team's approval.

derSascha:
> Thanks for the PR! I will consult my team about the issues that you raised. Specifically, "leaking information of their systems without asking them before".

This would be great, otherwise every installations makes every 10 minutes a HTTP call to the default S3 bucket.

> > Another very bad idea is to write into gem directories.
> 
> Could you elaborate more about that? What makes it "very bad"?

It's bad if you install gems as another user and the user running the code don't have write access to them or you share the bundle directory with different projects/environments. In this case, one project writes it configuration into the gem directory and the other uses them as default.

> > One of the tests does not like my e-mail address, the pull request includes a fix to support subdomains in the test.
> 
> If you open a new PR just with that change, I will merge it immediately. The rest of the code will have to wait for my team's approval.

The e-mail thing is not important and can wait. Thanks for the fast reply.

destari:
Thanks for the PR!  The remote settings can all be overridden by the user, and as you can see in the code we absolutely do not send any information that is sensitive in any way.  Our goal with this set of changes is to support upcoming features to allow our users to manage the configurations from a single point in the app, and manage a whole project, or fleet of notifiers, so they can maintain consistency across their platforms.  
The local file writes are something we will definitely change, thanks for the input on that!

As for disabling this mechanism altogether, I'd actually prefer a slightly different approach where we make the option only fetch on startup - which could be configured via the app in that case, and also solves an issue where someone disables it, forgets, leaves a company or whatever, and nobody knows why things are different for that system vs others.  We will discuss a bit more and put another reply here soon!

derSascha:
> as you can see in the code we absolutely do not send any information that is sensitive in any way

Its like a tracking pixel added to gem. Yes, it does not send highly sensitive information, but it sends the IP/OS/ruby version/gem version every 10 minutes of _every_ installation running the gem. This means local development setup, testing CI and production systems. Please consider if this heavy data collection is required?

> As for disabling this mechanism altogether, I'd actually prefer a slightly different approach

Any ideas are welcome, but please add some simple option to disabled it.

> We will discuss a bit more and put another reply here soon!

Thanks for the update!

destari:
> > as you can see in the code we absolutely do not send any information that is sensitive in any way
> 
> Its like a tracking pixel added to gem. Yes, it does not send highly sensitive information, but it sends the IP/OS/ruby version/gem version every 10 minutes of _every_ installation running the gem. This means local development setup, testing CI and production systems. Please consider if this heavy data collection is required?

I see your point here.  Couple things - the IP obviously is something that can be determined when the notifiers send the data they need to send anyway, so obviously that isn't related to this config option really at all.  The OS/Ruby version/etc are used so we can do contextual based configuration in the future, and we can also determine much better if notifiers are out of date, have security issues, etc, but I can understand the concern here. We definitely are not doing any user tracking or anything like a tracking pixel!


> 
> > As for disabling this mechanism altogether, I'd actually prefer a slightly different approach
> 
> Any ideas are welcome, but please add some simple option to disabled it.


Ok - we are having an internal discussion about the various approaches, and what would be the best option for our customers.

> 
> > We will discuss a bit more and put another reply here soon!
> 
> Thanks for the update!

kyrylo:
We decided that an option to disable notifier config functionality shouldn't be supported. I merged https://github.com/airbrake/airbrake-ruby/pull/622 as an alternative for test environments.

derSascha:
@kyrylo what about adding the disable notifier option, but do not enable it per default?

kyrylo:
We already have it (if I understood you correctly):
https://github.com/airbrake/airbrake-ruby/blob/88a67845f04860b8ee2cf6c4b371d314a3f9f2a1/lib/airbrake-ruby/config.rb#L121-L125

Disabling by default would be a breaking change and we don't want it.

derSascha:
That's the error notifications settings you mention, i'm talking about the remote config settings.

kyrylo:
Sorry, I don't see how this question is different from the one that I've already answered. The answer is still the same :) We won't add any new options.

derSascha:
@kyrylo sorry aout the misspelling. I was meant to talk about the remote configuration (not the notifier) and the question was: what about adding such option with default remote configs are fetched.

The PR changes three lines of code, not sure about what we are talking here? It's okay that you don't wan't add an option, but the PR is already here and and the work is done. We can change the default to `true` if this is the problem?

Otherwise please _explain_ why such an option is such a problem.

kyrylo:
We don't want to introduce this option because we want to be able to re-route the traffic that is coming to us to different servers (balance the load). If people can disable this feature, we no longer can do this reliably. In the future we might add new options to our UI and allow you configuring the notifier without restarting your server. Another concern is that the config already has a lot of options and the cost of adding a new option is high. It's not the reason but it is a reason.

derSascha:
We can add a warning around the option to encourage the users not to change this if they don't know what happens.

Sorry, but load regulation using text files on S3 buckets? Seriously? Sounds like a very odd excuse or a very bad decision.

derSascha:
Note: The requested option was merged by kyrylo in its own PR in #636.

Issue #626, opened 2020-11-05
Add GDPR infos about remote configuration


derSascha:
Hi,

every company running this gem together with airbrake has actually a contract with airbrake, including all required things about the GDPR. The new remote config option introduced in 5.0 runs on every developer notebook as well and this requires some more GDPR information to the developers. Please add them to the readme and add a way to decline the data collection, otherwise its impossible to use this gem anymore inside the European Union.

Thanks

destari:
Thanks for the comments!

Can you provide a bit more info on what you are expecting to see in our documentation? I am unclear on what exactly is missing, and how this is perhaps not already covered, so I want to be sure we are going the right direction from your point of view.

Also, the config feature is pulling a general config, by only using an HTTP GET request, and the only information that is passed (so we can correctly provide that configuration) are: notifier name and version, os, and language and version.  

This information is the same that is passed to our service when the notifier sends error information to us for your use.  The difference is that the config mechanism is designed so we can provide the correct configs without doing an HTTP POST or PUT, for example.

derSascha:
Thanks for the response!

To be clear, there is a difference where you actually run the gem:

1. On production server: Its expected that data gets transmitted to airbrake and the customers of airbrake has signed contracts.
2. On CI servers / testing farm: not expected that any data gets transmitted
3. On developer notebooks: not expected that any data gets transmitted and no contract or GDPR things signed

I'm talking about the last point here, the developer notebook.

> Can you provide a bit more info on what you are expecting to see in our documentation? I am unclear on what exactly is missing, and how this is perhaps not already covered, so I want to be sure we are going the right direction from your point of view.

The usual GDPR stuff, see [Wikipedia](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) (in short: what data gets collected, why it gets collected, how the data gets processed, how the user can request a dump of its data and howto request a removal of its data)

> Also, the config feature is pulling a general config, by only using an HTTP GET request, and the only information that is passed (so we can correctly provide that configuration) are: notifier name and version, os, and language and version.

And project identifier and IP address of the client...

> This information is the same that is passed to our service when the notifier sends error information to us for your use. The difference is that the config mechanism is designed so we can provide the correct configs without doing an HTTP POST or PUT, for example.

[...]

PR #632, opened 2020-11-26
Disable remote config if remote_config_host is nil

javierav:
This PR allow to developers to disable the remote config feature if set `remote_config_host` to `nil` in Airbrake configuration. Related with: #626


destari:
This looks good to me, I think we should take this PR, but I will let @kyrylo review officially.

kyrylo (approved):
This is an official review and this change looks officially good to me.

kyrylo:
After some internal discussion with my team we decided to revert this PR and introduce the `remote_config` option instead (https://github.com/airbrake/airbrake-ruby/pull/636). The new option is less hacky and more declarative. It will be available in v5.2.0.

PR #640, opened 2020-12-29
Add `remote_config_logging` option

frederikspang:
Adds a new configuration option `remote_config_logging` to avoid logging in certain situations, such as when running in Rails console (`defined?(Rails::Console)`) or similar.

This idea came from having long-running consoles and/or one-off jobs.

Other ideas:
- Might want to add proc/callable support for the option, to do things such as `config.remote_config_logging = ->{defined?(Rails::Console)}` or more advanced cases. Should run from app-server, but not from sidekiq/resque instance? etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -476,6 +476,22 @@ end
 Note: it is not recommended to disable this feature. It might negatively impact
 how your notifier works. Please use this option with caution.
 
+#### remote_config_logging
+
+Configures logging from the remote configuration feature. At intervals as
+described above, new settings will be fetched and output to the logger. These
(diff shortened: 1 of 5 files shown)


kyrylo:
We try not to introduce too many options in order to keep the config simple. The option that you suggest is very specific and it would be useful only in the scenarios that you described (long-running console processes or quick jobs). 

I am curious, why is this a problem? We use `debug` severity and the message that you're trying to silence is debugging information. Have you considered changing the Airbrake logger severity in development? Maybe that would be a simple fix in this situation.

PR #648, opened 2021-05-27
usefull add_filter example - exception message match 

equivalent:
it's obvious but will help tone of developers that are not aware that Exceptions in ruby can be matched as string

### DIFF
--- a/README.md
+++ b/README.md
@@ -590,6 +590,7 @@ Instead, you can ignore notices based on some condition.
 ```ruby
 Airbrake.add_filter do |notice|
   notice.ignore! if notice.stash[:exception].is_a?(StandardError)
+  notice.ignore! if notice.stash[:exception].message.match(/Couldn't find ActiveStorage::Blob with 'id'/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -590,6 +590,7 @@ Instead, you can ignore notices based on some condition.
 ```ruby
   notice.ignore! if notice.stash[:exception].is_a?(StandardError)
+  notice.ignore! if notice.stash[:exception].message.match(/Couldn't find ActiveStorage::Blob with 'id'/)
```
kyrylo, line 593: No objections to the change but I would prefer to split this into multiple lines since now this line too long (rubocop would complain).

equivalent: good point, I'll refactor this after we settle discussion https://github.com/airbrake/airbrake-ruby/pull/649#discussion_r640751906

equivalent:
closing this in favor of  https://github.com/airbrake/airbrake-ruby/pull/650

PR #649, opened 2021-05-27
usefull add_filter example - exception condition by ENV variable

equivalent:
for more inspirations for developers trying to avoid Airbrake notifying every message on every host

### DIFF
--- a/README.md
+++ b/README.md
@@ -590,6 +590,7 @@ Instead, you can ignore notices based on some condition.
 ```ruby
 Airbrake.add_filter do |notice|
   notice.ignore! if notice.stash[:exception].is_a?(StandardError)
+  notice.ignore! if notice.stash[:exception].is_a?(Net::ReadTimeout) if ENV['HOSTNAME'].match('storage-server')
 end
 ```
 


README.md
```diff
@@ -590,6 +590,7 @@ Instead, you can ignore notices based on some condition.
 ```ruby
   notice.ignore! if notice.stash[:exception].is_a?(StandardError)
+  notice.ignore! if notice.stash[:exception].is_a?(Net::ReadTimeout) if ENV['HOSTNAME'].match('storage-server')
```
kyrylo, line 593: No objections to the change but I would prefer to split this into multiple lines since now this line too long (rubocop would complain).

scottsbaldwin: Would it be better to do:

```
notice.ignore! if notice.stash[:exception].is_a?(Net::ReadTimeout) && ENV['HOSTNAME'].match('storage-server')
# or
notice.ignore! if ENV['HOSTNAME'].match('storage-server') && notice.stash[:exception].is_a?(Net::ReadTimeout)
```

equivalent: @kyrylo good point I'll refactor it tonight 

@scottsbaldwin  I see your point and I was thinking about it too (as that's the way how I use this examples in app as they are safer)  but those are combined examples of solutions. So different combinations of what can be done may not be as readable as single condition examples.

Now I don't mind either   ¯\_(ツ)_/¯ .

 If you guys tell me "let's provide safer way" => I'll refactor it to example Scott propose. If you tell me "readability of docs is priority" => let's keep the original refactored  to multiple lines 

now as we are refactoring this to multiple lines anyway I don't think it's that big of an issue to provide combination of conditions (what Scott is proposing)

kyrylo: Thanks for your elaboration. I think it would be better to refactor it to a more standard style. I thought the double `if` was a syntax error (it turned out to be valid Ruby). I think I won't be alone.

equivalent: no @kyrylo  this is me being stupid , yes this is type :sweat_smile:  ...accidentally it will work but it's a typo. Also 

@scottsbaldwin sorry I've misunderstood you due my typo, 

My bad sorry both

equivalent:
closing this in favor of #650  

In the end I've decided just to submit this one example as the example with ENV variables is ugly once limited to 80 chars per line (and will just confuse everyone)

PR #650, opened 2021-05-28
add_filter example to README

equivalent:
related to discussion in
* https://github.com/airbrake/airbrake-ruby/pull/649
* https://github.com/airbrake/airbrake-ruby/pull/648
@kyrylo 

In the end I've decided just to submit this one example as the example with ENV variables is ugly once limited to 80 chars per line (and will just confuse everyone)


kyrylo (approved):
Thanks!

PR #659, opened 2021-06-17
Implement  a thread-safe `Airbrake::Context`

### LINKED ISSUE #658: Is Airbrake.merge_context thread safe?
We are using Airbrake with shoryuken gem. We have noticed atleast one instance where the airbrake reported context is different from the one where error occured. Upon closer look, I find that we are relying on `Airbrake.merge_context` to setup the context before we process each shoryuken request. We are calling `Airbrake.notify` when error/exception occurs, which as I understand, is an async call by default.

Perusing the gem code seems to suggest that `Airbrake.notify` does the right thing of storing the context inside Notifier for async notification to pickup. However, I don't see any Thread.current in the call chain from Airbrake to NoticeNotifier etc. I wonder if this Airbrake code is threadsafe?

This documentation does not make it explicit: https://github.com/airbrake/airbrake-ruby#airbrakemerge_context

kyrylo:
Fixes #658 (Is Airbrake.merge_context thread safe?)

This makes sure any operations on the context hash are thread-safe. We do
synchronize calls around the code that attaches the context object to the error
report but the context object is not thread-safe itself.

### DIFF
--- a/spec/context_spec.rb
+++ b/spec/context_spec.rb
@@ -1,0 +1,54 @@
+RSpec.describe Airbrake::Context do
+  subject(:context) { described_class.current }
+
+  before { described_class.current.clear }
+
+  after { described_class.current.clear }
+
+  describe "#merge!" do
+    it "merges the given context with the current one" do
+      context.merge!(apples: 'oranges')
+      expect(context.to_h).to match(apples: 'oranges')
+    end
+  end
+
+  describe "#clear" do
+    it "clears the context" do
+      context.merge!(apples: 'oranges')
+      context.clear
+      expect(context.to_h).to be_empty
+    end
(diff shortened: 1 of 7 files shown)


spec/context_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  context "when another thread is spawned" do
+    it "doesn't clash with other threads' contexts" do
+      described_class.current.merge!(apples: 'oranges')
+      th = Thread.new do
+        described_class.current.merge!(foos: 'bars')
+      end
+      th.join
+      expect(described_class.current.to_h).to match(apples: 'oranges')
+    end
+  end
```
sjain, line 53: 👍

PR #673, opened 2021-12-21
Workaround in TimeTruncate with Rails 7.0.0

tnir:
`time` argument in `Airbrake::TimeTruncate` is now a Float in Rails 7.0.0 unlike an instance of `Time` in Rails 6.x or earlier.

Relates to airbrake/airbrake#1205

Signed-off-by: Takuya Noguchi <takninnovationresearch@gmail.com>

### DIFF
--- a/lib/airbrake-ruby/time_truncate.rb
+++ b/lib/airbrake-ruby/time_truncate.rb
@@ -9,7 +9,8 @@ module Airbrake
     # @param [Time] time
     # @return [String]
     def self.utc_truncate_minutes(time)
-      tm = time.getutc
+      # time does not have getutc method instead of Time on Rails 7.0.0
+      tm = time.respond_to?(:getutc) ? time.getutc : Time.at(time).getutc
 
       Time.utc(tm.year, tm.month, tm.day, tm.hour, tm.min).to_datetime.rfc3339
     end
--- a/spec/time_truncate_spec.rb
+++ b/spec/time_truncate_spec.rb
@@ -11,5 +11,11 @@ RSpec.describe Airbrake::TimeTruncate do
       expect(described_class.utc_truncate_minutes(time))
         .to eq('2018-01-01T05:00:00+00:00')
     end
+
+    it "converts it well even if the argument is an instance of Float" do
+      time = Time.new(2018, 1, 1, 0, 0, 20, '-05:00').to_f
+      expect(described_class.utc_truncate_minutes(time))
+        .to eq('2018-01-01T05:00:00+00:00')
+    end
   end
 end


tnir:
@kyrylo What do you think?

scottsbaldwin:
@thompiler can you also take a look at this?

thompiler:
>@thompiler can you also take a look at this?

Yes, I will take a look.

lib/airbrake-ruby/time_truncate.rb
```diff
@@ -9,7 +9,8 @@ module TimeTruncate
     # @param [Time] time
     # @return [String]
     def self.utc_truncate_minutes(time)
-      tm = time.getutc
+      # time is an instance of Float instead of Time on Rails 7.0.0
+      tm = time.is_a?(Float) ? Time.at(time).getutc : time.getutc
```
kyrylo, line 13: We should rather check for `getutc` existence than for the class of `time`. It would be more sturdy.

thompiler:
Looks generally good to me but could you incorporate @kyrylo's suggestion? You'll probably need to update the test/description to reflect the slight change in behavior.

lib/airbrake-ruby/time_truncate.rb
tnir: While I am not sure if it is better, fixed it in f4bdea4.

tnir:
@thompiler Thanks to pinging me. Back to you.

lib/airbrake-ruby/time_truncate.rb
kyrylo: Thanks! Curious to know why it’s not 🙂

lib/airbrake-ruby/time_truncate.rb
```diff
@@ -9,7 +9,8 @@ module TimeTruncate
     # @param [Time] time
     # @return [String]
     def self.utc_truncate_minutes(time)
-      tm = time.getutc
+      # time does not have getutc method instead of Time on Rails 7.0.0
```
thompiler, line 12: IMO, it might be clearer to remove the "instead of Time" bit. eg:
```suggestion
      # In Rails 7.0.0, time is a float and does not have the getutc method
```

thompiler (approved):
One [suggestion](https://github.com/airbrake/airbrake-ruby/pull/673/files#r779062856) on the comment. Other than that, LGTM.

lib/airbrake-ruby/time_truncate.rb
xiazek: even better : `..time is a Float..`

kyrylo (approved):

kyrylo:
@tnir thanks for the fix!

PR #676, opened 2022-01-10
time_truncate: simplify conversion code and tests

kyrylo:


tnir:
Thanks for updating and releasing a new version 6.0.2!

tnir:
@kyrylo Thanks.

By the way, why are we getting the gem with the dependency `rbtree3 ~> 0.5` while the gemspec specifies `~> 0.6` at https://github.com/airbrake/airbrake-ruby/blob/f72c428351b29defd4ae706e0c7e38bc842fa45f/airbrake-ruby.gemspec#L34 ?

https://rubygems.org/gems/airbrake-ruby/versions/6.0.2/dependencies also says it has the dependency `rbtree3 ~> 0.5`.

The following is what I got when updating airbrake-ruby 6.0.1 with #673 to 6.0.2 🤔 :
```diff
diff --git a/Gemfile b/Gemfile
index 0dd3d4ea7..81f91361a 100644
--- a/Gemfile
+++ b/Gemfile
@@ -173,6 +173,6 @@ gem "elasticsearch-rails", "7.2.0"
 gem "airbrake", github: "tnir/airbrake", ref: "01a551c0"
-gem "airbrake-ruby", github: "tnir/airbrake-ruby", ref: "f4bdea43"
+gem "airbrake-ruby", "~> 6.0.2"
diff --git a/Gemfile.lock b/Gemfile.lock
index fd2ebcf91..0ae11740c 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,11 +1,3 @@
-GIT
-  remote: https://github.com/tnir/airbrake-ruby.git
-  revision: f4bdea433af8b304f4b1e07574af83f5a3395cb8
-  ref: f4bdea43
-  specs:
-    airbrake-ruby (6.0.1)
-      rbtree3 (~> 0.6)
-
 GIT
   remote: https://github.com/tnir/airbrake.git
   revision: 01a551c01fd28b9188eddc3457710c8f9a0fddd4
@@ -88,6 +80,8 @@ GEM
+    airbrake-ruby (6.0.2)
+      rbtree3 (~> 0.5)
@@ -614,7 +608,7 @@ DEPENDENCIES
   airbrake!
-  airbrake-ruby!
+  airbrake-ruby (~> 6.0.2)
```

kyrylo:
Does `bundle update rbtree3` work? The `~> 0.6` specification is permissive. BTW: You want to be running rbtree3 0.7.0

tnir:
I just meant the difference between the source code on Git (GitHub) (`~> 0.6`) and built gem on Rubygem (`~> 0.5`). My Gemfile.lock does not matter in this discussion.

My point was the difference between https://github.com/airbrake/airbrake-ruby/blob/f72c428351b29defd4ae706e0c7e38bc842fa45f/airbrake-ruby.gemspec#L34 (`~> 0.6`) and https://rubygems.org/gems/airbrake-ruby/versions/6.0.2/dependencies (`rbtree3 ~> 0.5`).

kyrylo:
Understood. My guess is that RubyGems cached that version since it was the first version that we specified and it doesn't see the difference between `~> 0.6` or `~> 0.123`, since it's basically the same thing.

Issue #704, opened 2022-10-27
sanitize_hash in KeysFilter may leak sensitive blocklisted params


micred:
* Airbrake version: 6.2.0
* Ruby version: 3.1.0
* Framework name & version: Rails 7.0.2.3

#### Airbrake config
```
Airbrake.configure do |config|
  config.blocklist_keys = [:password]
end
```
#### Description
I trigger this notification:
```
Airbrake.notify('test',
   password: 'p',
   nested: {
     password: 'p'
   },
   openstruct: OpenStruct.new({password: 'p'}),
   params: ActionController::Parameters.new(password: 'p'), # Rails
   record: User.first, # Rails
   relation: User.all  # Rails
)
```

I expected to have the passwords filtered in any object, but only the first two were filtered.

I know that the user should cast the data to hash _before_ calling Airbrake.notify, but it is allowed to the user (no deep check on the values inside the hash, that may not be hash themself).
It leads to very dangerous sensitive data leaks of blocklisted attributes (like passwords in plain text).

kyrylo:
Hi, @micred! Thanks for opening this issue.

> I expected to have the passwords filtered in any object,

Where does this expectation come from? Do other error tracking tools do that?

> I know that the user should cast the data to hash before calling Airbrake.notify

From what I remember, I think our docs don't make this clear. But the behavior you demonstrated serves as proof of your statement. I think we should make this clearer in the docs as well.

micred:
> Where does this expectation come from? Do other error tracking tools do?

This expectation comes from the README:
> blocklist_keys - Specifies which keys in the payload (parameters, session data, environment data, etc) should be filtered.

Ok, no worries if you don't want to merge the PR, makes sense.

PR #705, opened 2022-10-27
sanitize_hash in KeysFilter to explicitly cast params to hashes

micred:
Preprocess any object to sanitize and explicitly cast their values into hashes.
It allows KeysFilter to filter them.
This PR contains framework-specific logic, but works on plain Ruby and can also include other framework-specific sanitizing methods to avoid dangerous sensitive data leaks of blocklisted attributes.

I understand if you prefer to keep away framework-specific logic from the gem. Still, in that case, I suggest blocking any object that is not a Hash to avoid automatic casting later, skipping the KeysFilter logic (KeysAllowlist and KeysBlocklist).

https://github.com/airbrake/airbrake-ruby/issues/704

### DIFF
--- a/lib/airbrake-ruby/filters/keys_filter.rb
+++ b/lib/airbrake-ruby/filters/keys_filter.rb
@@ -82,24 +82,46 @@ module Airbrake
 
       private
 
-      def filter_hash(hash) # rubocop:disable Metrics/AbcSize
-        return hash unless hash.is_a?(Hash)
+      def filter_hash object
+        object = sanitize_hash object
 
-        hash_copy = hash.dup
-
-        hash.each_key do |key|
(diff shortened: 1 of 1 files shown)


kyrylo:
1. Is there a way to write this in the form of an Airbrake filter?

    We don't want to tie `airbrake-ruby` to any integration, such as Rails and its `ActiveRecord::Relation` class. Therefore, this codebase is mostly free from that. The code should probably live here: https://github.com/airbrake/airbrake/tree/master/lib/airbrake/rack

2. We need to have tests for this critical functionality.

kyrylo:
This will degrade the performance of the notify function significantly. The solution is to serialize your objects prior to using it with `notify`.

PR #711, opened 2022-12-15
config: deprecate 'host'

### LINKED ISSUE #710: Set error_host and apm_host to host by default
Hallo @airbake team

Please set the default configuration settings https://github.com/airbrake/airbrake-ruby/blob/master/lib/airbrake-ruby/config.rb#L166-L167 to `host` and have maybe `host` pointing to `https://api.airbrake.io`. 

As it is configured  now, people using https://github.com/airbrake/airbrake and setting only `config.host` would send unnecessarily  their performance and (maybe error data?) to `airbrake.io`

kyrylo:
Fixes #710
(Set error_host and apm_host to host by default)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ### master
 
+- Deprecated the `host` option in favor of `error_host`
+  ([#711](https://github.com/airbrake/airbrake-ruby/pull/711))
+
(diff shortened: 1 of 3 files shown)


vpereira:
Thank you for your help and support @kyrylo

kyrylo:
Welcome 🍺

PR #721, opened 2023-07-07
Use #error_host instead of #host to avoid triggering a deprecation warning with deployment tracking

svenwin:
When using deployment tracking via `Airbrake.notify_deploy` we always get a deprecation warning:

    W, [2023-07-07T14:42:40.940088 #3280]  WARN -- : **Airbrake: the 'host' option is deprecated. Use 'error_host' instead

This commit uses `error_host` instead, so no deprecation warning is triggered.

### DIFF
--- a/lib/airbrake-ruby/deploy_notifier.rb
+++ b/lib/airbrake-ruby/deploy_notifier.rb
@@ -27,7 +27,7 @@ module Airbrake
       @sender.send(
         deploy_info,
         promise,
-        URI.join(@config.host, "api/v4/projects/#{@config.project_id}/deploys"),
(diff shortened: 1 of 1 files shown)


svenwin:
@mmcdaris is there any chance of merging this?

svenwin:
Anyone alive?

mmcdaris (approved):
Sorry this one fell of our radar. Thanks for the contribution, this will be included in the next release.

mmcdaris:
[airbrake-ruby v6.2.2 has been released](https://rubygems.org/gems/airbrake-ruby). Thanks again for the contribution @svenwin 🎉👍

PR #726, opened 2023-12-30
Add `base64` to runtime dependencies

mishina2228:
[`base64`](https://github.com/ruby/base64) will not be part of the default gems since Ruby 3.4.0. This change silences the following warning:
```
/home/runner/work/airbrake-ruby/airbrake-ruby/lib/airbrake-ruby/stat.rb:1: warning: base64 was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec.
```

### DIFF
--- a/airbrake-ruby.gemspec
+++ b/airbrake-ruby.gemspec
@@ -27,6 +27,8 @@ DESC
     'rubygems_mfa_required' => 'true',
   }
 
+  s.add_dependency 'base64', '~> 0.1'
+
   if defined?(JRuby)
     s.add_dependency 'rbtree-jruby', '~> 0.2'
   else


mishina2228:
@mmcdaris Sorry for direct ping, but airbrake gem will not work on Ruby 3.4+ unless you put `gem 'base64'` in Gemfile.

```
/home/runner/work/resurrected_god/resurrected_god/vendor/bundle/ruby/3.4.0+0/gems/airbrake-ruby-6.2.2/lib/airbrake-ruby/stat.rb:1: warning: base64 was loaded from the standard library, but is not part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec.
```

Any chance to review it?

